PR #20, opened 2021-02-17
v1.2.0

enarjord:
- bug fixes
- backtester now fetches necessary data from exchange, instead of user having to supply it manually


raven-kus:
@enarjord nice!

PR #26, opened 2021-03-04
add Requirements.txt

JohnKearney1:
Adds a version specific requirements list and updates setup command syntax.

### DIFF
--- a/.idea/inspectionProfiles/profiles_settings.xml
+++ b/.idea/inspectionProfiles/profiles_settings.xml
@@ -1,6 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <settings>
-    <option name="USE_PROJECT_PROFILE" value="false" />
-    <version value="1.0" />
-  </settings>
-</component>
--- a/.idea/inspectionProfiles/profiles_settings.xml
+++ b/.idea/inspectionProfiles/profiles_settings.xml
@@ -1,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <settings>
+    <option name="USE_PROJECT_PROFILE" value="false" />
+    <version value="1.0" />
+  </settings>
+</component>
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ requires python >= 3.8
 
 dependencies, install with pip:
 
-`python3 -m pip install matplotlib pandas websockets ccxt hjson numba`
+`pip install -r requirements.txt`
 
 discord
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,6 @@
+websockets==8.1
+pandas==1.2.2
+ccxt==1.42.18
+numpy==1.20.1
+hjson==3.0.2
+numba==0.52.0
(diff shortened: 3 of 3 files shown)


enarjord:
is it advisable to require specific versions of all the packages?

maybe better to require specific version only when a later version breaks the bot?

JohnKearney1:
I haven't explicitly tested the requirements.txt with a different version of python (I'm on 3.8.7 & 3.8.8). It COULD potentially cause issues but I highly doubt it. Requirements are generated based on the lowest "working" version of each package necessary.

Since the passivbot/main branch was merged the with numba code acceleration passivbot NO LONGER works on python 3.9.x. 

Minimum passivbot python version is 3.8.

This means compatibility range was determined using python 3.8.7. It may be worth making sure anyone with python 3.8.5 can use the requirements.

Generally, specificity in the dependencies prevents dependency updates or code changes from breaking the bot unexpectedly, so it's good form to include the version if you know it. 

For the time being, it's probably inconsequential which versions people use, and the requirements could simply be specified by name so long as they have the correct python.

enarjord:
so for minimum requirement it should be greater-than-or-equal-to >=, rather than equal == ?

websockets>=8.1
pandas>=1.2.2
ccxt>=1.42.18
numpy>=1.20.1
hjson>=3.0.2
numba>=0.52.0

i tried installing a package with a lower version than current

having installed trio 0.18.0, using python3 -m pip install trio==0.17.0, it uninstalled 0.18.0 and installed 0.17.0

so using ==x.x.x means that when people run this, it'll downgrade their installed versions if their already installed version is higher.  which isn't a problem if people use docker or virtual env, which i guess is recommended practice in any case.

> For the time being, it's probably inconsequential which versions people use, and the requirements could simply be specified by name so long as they have the correct python.

as you say.  as far as i know, only python version needs to be 3.8, and all the other packages can be latest version

JohnKearney1:
> so for minimum requirement it should be greater-than-or-equal-to >=, rather than equal == ?

Correct. With a caveat;
If passiv were ever to move to a distribution environment (pypi for 'pip install passivbot') it would be best practice to specify exact package versions to prevent ambiguity.

For instance, passivbot uses numba v0.52.0 currently which only supports python 3.8. If the numba team chooses to update the package to support python 3.9, and any of the syntax also changed, the code for passivbot would break. Specifying exact versions ensures that updates to the DEPENDANCIES that are installed by the user don't break passivbot itself. 

If you can ensure forward support for all of the dependancies (up to their latest version) it's probably best to use '>=' as it doesn't uninstall newer versions of the same package. 


- Use '>=' if you can ensure passivbot always works with the latest ver of all dependancies.

- Use '==' if not, or if discontinuing forward support for a package. 

Keep in mind Requirements.txt is required for packaging with a setup.py. I don't see any obvious rush to package it up, but using a requirements.txt removes the need to troubleshoot errors for multiple versions of a dep (assuming only one version is allowed).

enarjord:
alright i suggest the following

because of what i read here https://pip.pypa.io/en/stable/user_guide/
we can use this syntax `pkg>=1.0,<=2.0` in case any package's newest version ever breaks passivbot.
so we leave all `pkg>=x.x.x` for now, except python itself which must be `python>=3.8.0,<3.9.0`

PR #34, opened 2021-03-26
Add missing aiomultiprocess dependency

moafred:


enarjord:
aiomultiprocess isn't used in v3.2 so better remove the import than add the dependency

moafred:
> aiomultiprocess isn't used in v3.2 so better remove the import than add the dependency

Okay, I think i had an issue while pulling [changes](https://github.com/enarjord/passivbot_futures/commit/069a20d8736aef1240a7f953b0ab8aa5ec12aba8).

PR #41, opened 2021-03-31
Ideas

moafred:
In this PR:
- docker stuff is about infrastructure
- logs are data
- requirements.txt contains the vendor and should be keep at the root
- changelog.txt should be at the root with readme.md, refactoring.md and so on.

----

**Api keys**
I think they should be in one file at the root of the repository, like: `user_config.example.json` with binance and bybit. This file could be given to the bot in the command line.

**Live settings file**
The magic behind getting the same file name in the live settings distub me, it could also be given through the CLI.

**Makefile**

Makefile is good way to give an unique entry point and durable contract for the user (like the docker-compose.yml file). It could be an idea to implement it like this:

```bash
.DEFAULT_GOAL: help

# List commands
help:
	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m * %s\033[0m %s\n", $$1, $$2}'

run-live: ## Run bot with live settings. make run-live EXCHANGE=binance USERNAME=username
	python3 start_bot.py $(EXCHANGE) $(USERNAME)

run-live: ## Run bot with live settings. make run-live CONFIG=configname
	python3 backtest.py $(CONFIG)
```


README.md
```diff
@@ -15,12 +15,14 @@ NOTHING here is final.
...
-> ├── `Data/`  
+> ├── `requirements.txt`
+> │── `changelog.txt`  
+> ├── `infrastructure/`  
 > │    ├── `Dockerfile`  
 > │    ├── `docker-compose.yml`  
-> │    ├── `changelog.txt`  
-> │    ├── `requirements.txt`  
+> ├── `data/`  
 > │    └── `historical_data/`
```
JohnKearney1: We might also want to think about moving `historical_data/` into the `bot/` directory:

1. If @enarjord is confident that the type of historical data used, naming scheme etc. will never change, we can leave it outside of the scope of the bot, so that if multiple versions (with different operating logic) were added to a the single PassivBbot package (ex. v2.1 and v3 allowing for hedge and cross mode simultaneously), they can pull from the same historical data. Also might make things easier for implementing a downloader. 

**Alternatively**

2. If there is any potential that the structure of historical data might change in future versions, we should move  `historical_data/` into the `bot/` directory. This will lessen confusion between versions, but might make getting historical data a slower process as it will be a per-version task. Potentially, we could have the data downloader reference these subfolders anyhow, assuming the format doesn't change. This seems like it may be the safer option.

JohnKearney1: Edit: This issue can be solved by placing historical data under `Data/historical_data/{version}`.

This was also unrelated to the PR. My bad!

JohnKearney1 (commented):
These changes address points 1 and 3 by moafred.

- Docker files were edited and moved into the `infrastructure/` directory.
- `requirements.txt` was moved to the root, and the example `setup.py` updated.

Not ready to merge, working on an API keyfile proposal to simplify users.

JohnKearney1 (commented):
Consolidates API keys into a single JSON file `api-keys-example.json`. 

Isolated test for reading the JSON is attached: [test.zip](https://github.com/enarjord/passivbot_futures/files/6241013/test.zip)

The new JSON is a single file, where the top level objects are the supported exchanges, and objects within that are the users. The load_key_secret() method was re-written to search for the given user `sys.argv[1]` within the JSON object named after the given exchange `sys.argv[2]`, then return the corresponding list. The keyfile is now explicitly defined.

README.md
moafred: You see the point ^^ each change can have multiple solution and so discussions, we cannot do the refactoring in only one PR.
I'm not comfortable if the `version` path because the schema of historical data may not changed between multiple version.
The best way for me is to not change anything until we have an issue about it.

moafred:
@JohnKearney1 i suggest you to create an issue with a proposal written in it for each topics:
- one file for api keys
- ..

Let's discuss in those issues and then create a branch/pr to make the proposal when we are aligned.

README.md
JohnKearney1: Got some feedback from @enarjord, will move each of these issues into separate threads.

JohnKearney1 (approved):
Rolled back API key changes and example `Setup.py` addition, but left dockerfile moves. 
As per @moafred suggestion, changes for each major segment of the refactor will be opened as an issue then approved / denied via pull request. 

What else is required to satiate this pull req?

README.md
```diff
@@ -6,21 +6,24 @@ Since this is such a broad and somewhat ambiguous process, there is no single co
 Thus, it's advantageous to get as much developer feedback as possible to ensure that the updated structure is easy to change, compartmentalize, test etc.
```
moafred, line 6: All this stuff added for the "refactor" should be move to the description of an issue rather that modifying the README.md of the project :)

infrastructure/docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   passivbot:
-    build: .
+    build: ..
```
moafred, line 3: The volume will be breaked because we changed the directory of docker-compose.yml file.
Maybe it's not worth to change the location of Dockerfile and docker-compose.yml, it add complexity in the resolution of the paths for only two files.

PR #43, opened 2021-04-01
Consolidate API Keys

JohnKearney1:
# API Keys - Consolidating Into Single Keyfile 

This PR focuses on updating the API key logic to use a single, easily configurable JSON file. This PR implies a change in the file structure, but is not major. This is one of multiple changes within the refactoring process.

___

**Currently**, API Keys are fetched from a found file `api_key_secrets/{exchange}/{user}.json` where `exchange`  is `sys.argv[1]` and `user` is `sys.argv[2]`. 

Under the current model, a sub-folder for each exchange is necessary, and the user must configure multiple, separate JSON files to add a new API key pair. The exchange input is also case sensitive, and must match the name of the exchange's folder. While searching a directory for any number of variable filenames works, referencing a dictionary eliminates the need for excessive formatting within the code, decreases the complexity of the file structure, and makes future integrations easier.

**With this PR**, API Keys are fetched from the `api-keys.json` using the same exchange and user arguments to search the dictionary for matching entries. The result of the search is returned as a dictionary that can be indexed to parse the key vs. the secret key.

This solution allows future integrations that utilize API Keys (namely a GUI / CLI interface) to reference a single file and easily add, remove, or edit entries using the template keyfile's given structure.

**Changes**  
- Add `api-keys.json`
- Edited `passivbot.py`
- Removed `api_key_secrets/`, it's subfolders and contents.

**Tasks**
- [x] Create template keyfile (`api-keys.json`)
- [x] Edit `passivbot.py` to support keyfile
- [x] Ensure multiple instances can access the file simultaneously
- [ ] Review template keyfile structure
- [ ] Review implementation
- [ ] Make changes if necessary
- [ ] Ensure compatibility with integrations

**Isolated Test Case**
[Isolated Test.zip](https://github.com/enarjord/passivbot_futures/files/6245060/Isolated.Test.zip)

**Addl.**
- Updated .gitignore for PyCharm IDE


api-keys.json
```diff
@@ -0,0 +1,14 @@
...
+    "User2" : ["API_KEY", "API_KEY_SECRET"]
+  },
+
+  "BYBIT" :
+  {
+    "User1" : ["API_KEY", "API_KEY_SECRET"],
+    "User2" : ["API_KEY", "API_KEY_SECRET"]
+  }
+
+}
```
JohnKearney1: JSON structure here is a suggestion, but should be reviewed by other contributors to ensure it's easy to integrate with existing and future work. This structure was proposed as it requires very minimal changes to `passivbot.py`.

moafred: Nice to have one place for the secrets :)
In your proposition, i suggest to keep the binance and bybit in lowercase and change from array to object the api key and secret like this:
```
    "User1" : {
        "key": "key",
        "secret": "secret"
    },`
```

I have another proposition of the structure:
Define accounts and put the exchange in it. It will change the arguments cli by giving the account name instead of the exchange + username, like this `python3 start_bot.py account_binance_1 binance/xmr.json`.
```
{
    "account_binance_1" : {
        "exchange": "binance",
        "key": "key",
        "secret": "secret"
    },
    "account_binance_2" : {
        "exchange": "binance",
        "key": "key",
        "secret": "secret"
    },
    "account_bybit_1" : {
        "exchange": "bybit",
        "key": "key",
        "secret": "secret"
    }
}
```

enarjord: worth keeping in mind possible upcoming expansions to new futures exchanges like ftx, kraken, deribit etc., as well as non futures markets like spot and margin.  it should be made easy to separate, if support for them is added, binance BTCUSDT futures, BTCUSDT spot, cross margin, isolated margin

jontstaz: I was thinking of leveraging a SQLite3 DB to hold users. Column structure could be something like [id, api_key, secret_key, exchange, config]

Snippet below. Would give user the option to either use the central JSON file or an sqlite db.

```
dbEnabled = False # Use sqlite3 (True) or Json (False)
accountID = 231 #this would be fetched from initial start_bot.py parameter (eg. --accid=231)

if dbEnabled == False:
    # Fetch keys from json file
    exchange = sys.argv[1]
    user = sys.argv[2]
    p = Popen(f"python3 {exchange}.py False {user} --nojit", shell=True)
    print(f"\nStarting {exchange} {user}")
else:        
    conn = sqlite3.connect('passivdb.sql')
    c = conn.cursor()
    c.execute("SELECT * FROM {tn} WHERE {idf}={my_id}".\
            format(tn='users', idf='id', my_id=accountID))
    userInfo = c.fetchone()
    if userInfo:
        exchange = userInfo[3]
        user = userInfo[4] #can reuse same config for multiple users
        p = Popen(f"python3 {exchange}.py True {user} --nojit", shell=True)
        print(f"\nStarting {exchange} for ID {user} in DB")
    else:
        print("User ID not found in DB")
```

moafred: Users doesn't change a lot and i think it add complexity for insert/read/migrate data in a db versus a file. My point of view :)

> worth keeping in mind possible upcoming expansions to new futures exchanges like ftx, kraken, deribit etc., as well as non futures markets like spot and margin. it should be made easy to separate, if support for them is added, binance BTCUSDT futures, BTCUSDT spot, cross margin, isolated margin

So, maybe we need a new key like `market` for each user ? (It's why i prefer to have all the information into the account instead having the exchange at the root level)

jontstaz: Yeah agreed that it adds complexity. It's just my personal preference to use a DB to store data, especially sensitive data like API secrets which I'd store in a DB in a hashed format.

Another thing is I believe configs should be universally applicable, meaning multiple user can leverage the same config files. The centralised JSON file of users should have extra fields to specify the exchange and the config to use. So when starting a bot you would just specify the user via a parameter and then the exchange and config are automatically populated.

moafred: I think the config is not related to the user, so it's make sense for me to give to the bot the WHO (user on which exchange / market) and the WHAT (the strategy) (like in my first comment), what do you think ?

Concerning the sensitive data, i think all should have an encrypted partition ^^ use filesystem or DB is the same considering this, it's how we encrypt the data that matter in this case.
I'm not against storing the user config in the database, it's just for me it's more complicated for the end user to configure the bot and to handle migration of the db :)

JohnKearney1: > So, maybe we need a new key like `market` for each user?

I've added this in the commit I'm about to push, for the time being I've just filled in strings, but in the future, if the bot were ever to read the `market` key, we should probably change this to an int, to avoid string formatting, Example:

> market = 0 # Perpetual Futures,  
> market = 1 # Inverse Perpetual Futures,  
> market = 2 # Spot Trading (no leverage)  

and so on...

Since there's no implementation for it now, the example values are irrelevant, but before release we might consider establishing some type of standard structure.

JohnKearney1 (commented):
- Updated `api-keys.json` structure as per suggestion by @moafred  
   - Added `market` key
   - Changed "API_KEY" and "API_SECRET" to individual keys instead of string dict.
 - Updated `passivbot.py` logic accordingly
 
 **Isolated Test:** [Isolated Test (2).zip](https://github.com/enarjord/passivbot_futures/files/6250973/Isolated.Test.2.zip)

api-keys.json
JohnKearney1: > So when starting a bot you would just specify the user via a parameter and then the exchange and config are automatically populated.

This seems just to push the issue in a different direction in my opinion. If a given API Key is only linked with a single configuration, the user has to then edit the API keyfile every time they want to change configurations. Minor issue, sure, but imagine some type of interface having to duplicate your API keys for every configuration you add to the bot. The keyfile becomes messy, long, and the UI then has to be in charge of managing this keyfile. 

I have little experience with SQL Databases so I won't comment on that, except to say it seems like this should be leveraged along with the UI/UX integrations assuming they come to fruition.

enarjord: at the moment "symbol" and "exchange" are specified in config

maybe they shouldn't be because the same config can be used by different exchanges with different symbols

for example i use the same LTCUSDT config for both binance and bybit.  different user, different exchange, same symbol, same config

i also used an xmr config to test TRXUSD_PERP coin margined on binance.  same user, same exchange, different symbol, same config

bot needs to know
- api key/secret (user), which is bound to an exchange
- config
- symbol
- (maybe also specify market, in case of future implementation of spot and margin markets, where symbol is the same as in futures markets)

so
`python3 start_bot.py {user} {symbol} {path_to_config}`
which admittedly requires one more piece of info user has to specify when launching the bot

another benefit of pulling "symbol" and "exchange" out of the config is that configs become leaner and more obviously universally usable across markets

enarjord: i use same user for binance XMRUSDT and TRXUSD_PERP

and same user for bybit BTCUSDU21 and LTCUSDT

so exchange should not be bundled with api key/secret, lest there be duplications

JohnKearney1: > - api key/secret (user), which is bound to an exchange
> - (maybe also specify market, in case of future implementation of spot and margin markets, where symbol is the same as in futures markets)

These two are addressed [here.](https://github.com/enarjord/passivbot_futures/commit/1d1ccae4ae7892f35a09ffc62ccda4e524730b32)

In regards to market param, some definite structure will be necessary in the future, if used. See [this](https://github.com/enarjord/passivbot_futures/pull/43#discussion_r606379318) previous comment.

enarjord: market type (futures/spot/margin), in cases where symbol is identical but market is different, should not be bundled with the api key/secret for the exchange because if using, say, both BTCUSDT futures and BTCUSDT spot, that would necessitate two accounts with identical exchange and key/secret, but different market type

```
{
    "account_binance_1" : {
        "exchange": "binance",
        "key": "key",
        "secret": "secret"
    },
    "account_binance_2" : {
        "exchange": "binance",
        "key": "key",
        "secret": "secret"
    },
    "account_bybit_1" : {
        "exchange": "bybit",
        "key": "key",
        "secret": "secret"
    }
} 
```

just like this is enuf

i think market type should not be specified in live_config.json file either because they are much prettier when lean and universal, being agnostic about symbol, exchange, market type and other particulars

if distinguishing market types becomes necessary somewhere down the road, it could be done with an arg when launching start_bot:
`python3 start_bot.py user_name BTCUSDT path/to/live_config.json --market-type spot|futures|isolated-margin|cross-margin`

if unspecified, it defaults to `--market-type futures`

passivbot.py
```diff
@@ -810,11 +810,23 @@ def make_get_filepath(filepath: str) -> str:
...
-        return json.load(open(f'api_key_secrets/{exchange}/{user}.json'))
-    except(FileNotFoundError):
-        print(f'\n\nPlease specify {exchange} API key/secret in file\n\napi_key_secre' + \
-              f'ts/{exchange}/{user}.json\n\nformatted thus:\n["Ktnks95U...", "yDKRQqA6..."]\n\n')
-        raise Exception('api key secret missing')
+        keyfile = json.load(open('api-keys.json'))
+        # Checks that the user exists, and it is for the correct exchange
+        if user in keyfile and keyfile[user]["exchange"] == exchange:
+
+            # If we need to get the `market` key:
```
moafred, line 817: It will confuse more than it will give guidance to speak about a configurable market that is not yeat configurable, commented code is "dead code" and didn't give any information of the intention :)
What i suggest you is to remove the lines in api-keys.json and add a comment in this file with a link of a message from the discussion of this PR, if we need to add the feature of a configurable market we will inevitably see this comment left in the code :)

PR #50, opened 2021-04-09
Use argparse for parameterizing the options.

pbosch:
The arguments for backtester, downloader, and start_bot are better parameterized and offer some description in form of help text.

### DIFF
--- a/backtest.py
+++ b/backtest.py
@@ -1,23 +1,13 @@
-import sys
-import json
-import hjson
-import numpy as np
-import pandas as pd
-import asyncio
-import os
-import gc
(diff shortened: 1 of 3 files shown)


enarjord:
i suggest adding support for positional args, in addition to -a/--arg syntax

also, when launching backtest.py backtest config's whole path should be specified.  currently, it is `python3 backtest.py xmr` and behind the scenes it looks in dir backtest_configs/ and appends .hjson

enarjord:
argparse now implemented, this old PR no longer needed as reminder

PR #53, opened 2021-04-11
bugfix: fetch_ticks_time was missing from bybit

dclausen:


pbosch:
The last time I checked, Bybit didn't support fetching based on time. This is also the reason why the find_time function in the downloader searches for the specified start time first instead of just fetching it directly like on Binance.

enarjord:
pbosch's downloader tries fetch_ticks_time first, if it throws an exception (fetch_ticks_time being undefined for bybit) it'll fetch_ticks instead and search for start time

if fetch_ticks_time is added to bybit, this logic will fail

PR #74, opened 2021-04-25
Proposal: Improve indexing

pbosch:
Improve indexing by using an enum to track the index of the parameter array. The index needs to be cast with np.intp to work with numba. It aims to make the code more readable and maintainable. The performance is the same as with direct access to the tuple. Also replaces the tuple with a numpy array to allow booleans in it. As numpy arrays only support single data types, those are cast to float, but we can still check against them.


enarjord:
found another way using globally defined Enum, which also yielded an unexpected speed improvement

PR #80, opened 2021-05-02
Add git as package for docker

moafred:


migueltg (commented):
Can you put in same layer? In same Run that pip install to improve performance

enarjord:
merging this one, if it should be put in same layer as migueltg suggests, make another PR, migueltg or moafred

moafred:
> Can you put in same layer? In same Run that pip install to improve performance

It's not about performance but build time :)
The install of git should not change during time, it's why i put it in a dedicated layer at the top.
Below layers (copy and pip install) will be rebuilt more often due to change in the application.
So with a separate layer at the top, the install of git should never be rebuilt (except change of lines aboves).

PR #151, opened 2021-06-15
Update jupyter lab install

dclausen:


enarjord:
thx for the catch, jupyter lab is installed with `pip install jupyterlab`, `pip install jupyter-lab` won't work

for opening after installation, both `jupyter-lab` and `jupyter lab` works

PR #174, opened 2021-07-04
Add Sharp gain ratio in tuning

556isback:
Implementing sharp_gain_ratio in ray tuning aiming to smooth the pnl plot and prevent overfitting the dump in May 19th.

### DIFF
--- a/configs/optimize/default.hjson
+++ b/configs/optimize/default.hjson
@@ -8,6 +8,7 @@
   # set to 0.0 to disable breaking early
   break_early_factor: 0.5
 
+  minimum_sharp_gain_ratio: 1.2
   minimum_bankruptcy_distance: 0.5
   minimum_equity_balance_ratio: 0.5
   minimum_slice_adg: 1.0
--- a/configs/optimize/default.hjson
+++ b/configs/optimize/default.hjson
@@ -8,7 +8,8 @@
   # set to 0.0 to disable breaking early
   break_early_factor: 0.5
 
-  minimum_sharpe_gain_ratio: 1.2
+  minimum_sharpe_ratio: 1.2
+  sharpe_ratio_n_days: 7.0
   minimum_bankruptcy_distance: 0.5
   minimum_equity_balance_ratio: 0.5
   minimum_slice_adg: 1.0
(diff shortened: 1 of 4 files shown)


556isback:
Here is the plots after implementing the feature
balance_and_equity
pnl_cumsum_long
pnl_cumsum_shrt

enarjord:
from what i gather sharpe ratio is the ratio of mean(gain) to std(gain) for some chosen sampling

say chosen sample size is daily and daily pnls are [1.0, 1.5, 0.3, 0.0, 1.4, 1.2, -0.3, 3.0]
mean is 1.0125 and std is ~0.9765
so sharpe ratio is mean(pnls) / std(pnls) ~= 1.03684

have i understood this correctly?

enarjord:
fixed typos
and removed import from external lib empyrical,
instead calculating sharpe ratio locally by taking mean(daily_pnls) / std(daily_pnls),
making sure to fill days without pnl with zeros

556isback:
> from what i gather sharpe ratio is the ratio of mean(gain) to std(gain) for some chosen sampling
> 
> say chosen sample size is daily and daily pnls are [1.0, 1.5, 0.3, 0.0, 1.4, 1.2, -0.3, 3.0]
> mean is 1.0125 and std is ~0.9765
> so sharpe ratio is mean(pnls) / std(pnls) ~= 1.03684
> 
> have i understood this correctly?

Sounds about right, forgive me that I didn't dig in to the formula of sharp ratio.
But you have misunderstood my modification, what I mean by sharp_gain_ratio is (sharp ratio) / gain. Cause originally as I implementing the sharp ratio, I realize that the larger the gain the larger the sharp ratio, but the plots  might not be smooth, and on the contrary when the gain is low even sharp ratio is small but the plots might be very smooth. 
So I came up the idea of sharp gain ratio, which is  mean(pnls) / std(pnls) / (adg**n_days).

enarjord:
gotcha, changed it back
i'll have to experiment with it for myself a bit
feel free to elaborate more on what you have discovered

556isback:
Second thought, we just need to minimize standard deviation of the daily gain percentage to smooth the plot.

enarjord:
right

so a simple sharpe ratio on the percentage daily gains then

pnls.mean() / pnls.std()

where pnls is [day.pnl.sum() / day.balance.last() for day in fills.groupby(days)].reindex(all_days).fillna(0.0)

when using percentage gains instead of absolute gains, compounding profits makes no difference on the daily % gain, so i don't see the need to divide by total gain

PR #189, opened 2021-07-07
Pin requirements, remove not general packages

DmytroLitvinov:

PR #196, opened 2021-07-16
Update requirements.txt

RustyMarx:
Will now work in python 3.9


migueltg:
It's better fix version to prevent new further updates

DmytroLitvinov:
Can be closed in favour of https://github.com/enarjord/passivbot/pull/189

enarjord:
@hoeckxer that's my impression from what others have reported, but i haven't tested myself whether ray > v1.2 works properly

in my opinion it is desirable to upgrade ray such that python v3.9 may be used

PR #208, opened 2021-08-03
Docs additions

wojtekjakubo:
Adding simple guide for using multiple instances of the bot in the configuration.md

Adding warning in telegram.md for not using same telegram tokens on multiple configs.

Added specific explanation for which config file to use in optimization.md  process as a start point with -t flag. (.json)


rayures:
Great pointers! was just setting up a test with 2 bots on 1 exchange :-)

PR #287, opened 2021-11-16
Replace all the ``type()`` checks with ``isinstance()``

s0undt3ch:


enarjord (approved):
a quick duckduckgoing reveals isinstance() is preferred over type() because isinstance() supports inheritance

so no objections from me to using isinstance()

PR #288, opened 2021-11-16
Implement an `HTTPClient` to implement separation of concerns

s0undt3ch:


enarjord:
beautiful stuff

it fails for me when trying it live tho

for example line 38 in httpclient.py should be
`async with aiohttp.ClientSession().get(url) as response:`
not
`with aiohttp.ClientSession().get(url) as response:`

and other errors

do you prefer to merge it first and fix errors later?  or fix them first and merge after?

s0undt3ch:
Yeah, I fixed it locally but havent pushed it yet. Will do it in the morning and then I'll ping you.

PR #289, opened 2021-11-19
Switch from ``print()`` and ``print_()`` to using python's logging

### LINKED ISSUE #276: use logging module instead of custom `print_()`

s0undt3ch:
Fixes #276


enarjord:
works beautifully now

PR #290, opened 2021-11-19
`--nojit` is still respected while allowing `njit` to be used live

s0undt3ch:


s0undt3ch:
Need to fix the tests before merging.

s0undt3ch:
Should be good now.

enarjord (approved):
after one small bug fix, live now works both with and without numba

PR #294, opened 2021-11-23
Minimum python version is 3.8, let's import ``annotations`` from ``__future__``

s0undt3ch:


enarjord (approved):
tested live and works
looks neater this way

PR #295, opened 2021-11-24
Data structures instead of unconstrained dictionaries

s0undt3ch:


s0undt3ch:
Converting to draft since it will break `backtest`

PR #302, opened 2021-12-10
Only print output for enabled modes

s0undt3ch:


enarjord:
i suggest checking for pos and not checking if long/short is enabled in config

for example, if graceful stop is enabled on short, config will say short enabled false, but there may still be a short pos

enarjord (approved):

PR #306, opened 2021-12-27
removed telegram functions from passivbot core

LordChunder:
-removed telegram implementation from passivbot.py
-removed python telegram bot references from requirements.txt
-updated api-keys example to reflect removal of telegram bot
-left telegram_bot.py but renamed to telegram_bot_old.py

passivbot.py optimize.py and backtest.py all tested and working

### DIFF
--- a/api-keys.example.json
+++ b/api-keys.example.json
@@ -2,14 +2,7 @@
     "binance_01" : {
         "exchange": "binance",
         "key": "key",
-        "secret": "secret",
-        "telegram": {
-            "enabled": false,
-            "notify_entry_fill": true,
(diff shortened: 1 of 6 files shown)


goldmang:
It would be nice to have a telegram integration to check the status of all futures trades.

rayures:
> It would be nice to have a telegram integration to check the status of all futures trades.

when the goal is ' check status of all futures trades' then there are dashboards available to attain that goal.

- https://github.com/hoeckxer/exchanges_dashboard
- https://github.com/SH-Stark/trading-dashboard
- https://github.com/ecoppen/futuresboard

PR #322, opened 2022-01-20
Add Standard deviation as new metric

neuhausj:


enarjord:
already merged, thanks for the idea of using std on price action distance

PR #336, opened 2022-03-28
Allow the user to override the location of the api-keys.json file

hegjon:
The default is the same as before, so that the end-user experience is as the same

### DIFF
--- a/passivbot.py
+++ b/passivbot.py
@@ -1078,7 +1078,7 @@ async def main() -> None:
 
     args = parser.parse_args()
     try:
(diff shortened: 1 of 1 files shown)


passivbot.py
```diff
@@ -1078,7 +1078,7 @@ async def main() -> None:
 
     args = parser.parse_args()
     try:
-        accounts = json.load(config["api_keys"])
+        accounts = json.load(args.api_keys)
```
hegjon, line 1081: Thanks for spotting that one

enarjord:
tested and works

PR #354, opened 2022-08-27
Fixing api keys location when given as command line argument

amit-sides:
I have tried to run the bot with the `-ak` parameter, and I got an error saying the API Keys file is missing.
I noticed that in the `load_exchange_key_secret` function the location of the file is a constant and not updated according to the `-ak` flag.
Here is a backwards compatible fix.

P.S:
I also closed the `api-keys.json` file after it was read because there is no reason for it to remain open during the execution of the bot.

### DIFF
--- a/passivbot.py
+++ b/passivbot.py
@@ -107,7 +107,7 @@ class Bot:
 
         self.log_filepath = make_get_filepath(f"logs/{self.exchange}/{config['config_name']}.log")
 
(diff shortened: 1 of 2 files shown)


enarjord:
Many thanks for the PR

Made some more changes, simplifying a little.  Figured it was unnecessary to let args.api_keys be of type "open" in the first place, necessitating a close()

PR #358, opened 2022-11-10
Removed error when trying to cancel non existing order on Bybit. Adde…

kikoseijo:
Suppressed an error when trying to cancel a non existing order on Bybit exchange. 
Added exchange_leverage variable for convenience.

### DIFF
--- a/bybit.py
+++ b/bybit.py
@@ -16,6 +16,8 @@ from passivbot import Bot
 from procedures import print_async_exception, print_
 from pure_funcs import ts_to_date, sort_dict_keys, date_to_ts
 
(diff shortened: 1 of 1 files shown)


enarjord:
Thanks for the PR

instead of modifying it and merging, i pushed relevant changes to v5.8.1 branch:
- set leverage by kwarg instead of modifying python source code
- instead of suppression, less verbose error message when trying to cancel non-existing order

by returning empty dict instead of letting the exception block be executed, this code `self.ts_released["force_update"] = 0.0` would not be executed.  Its purpose is to force REST API update of open orders and position in case of error

PR #379, opened 2023-02-12
added KuCoin Exchange

migueltg:
First approach to add **KuCoin** to passivbot using API calls natively and websocket connection.
It has been tested for a week with various symbols but it does not mean that strange behaviors cannot occur.
Try with a small quantity and at your own risk.

Pending to test ohlcv mode


enarjord:
ohlcv mode works now
remaining to fully test and verify working websockets

PR #396, opened 2023-04-26
docs(forager): set ohlcv_interval to 1h

raftheunis87:


raftheunis87:
Nevermind, it is only 1h in the default forager config. If you don't specify a value, it will default to 15m.

PR #413, opened 2023-08-17
Filter part of the symbols according to minimum days on market

Mengflz:
Hi, thanks for your work. After running the forager.py for a relatively long period(~ 3 months). I noticed that it is easy to stuck into a situation, which is the rapid declines shortly after new currency pairs getting into the market, especially in long positions. So I try to add a parameter to filter out the currency pairs that do not meet the criteria.
It is not a perfect implement. Feel free to modify it. I am glad to hear any feedback on this.


enarjord:
thanks for the PR

closing this because the same functionality is already added to forager

PR #431, opened 2024-01-07
add venv var

GoempieK:
Adds venv var to forager config file


enarjord:
Thanks for the PR. Made some changes to it.

GoempieK:
Thanks i'm not really a programmer :-)

PR #453, opened 2024-06-20
upgrade Dockerfile to python 3.10

raftheunis87:


enarjord:
what's the difference between using python3.9 and python3.10 here?

raftheunis87:
I think it makes sense to use the latest working/supported version at all times? Python 3.9 will be deprecated before 3.10...

PR #510, opened 2025-06-15
Master

msei99:
bugfix fetch pnls bitget on a new api


enarjord:
with this change, the loop would break on first iter every time because end_id == all_data[-1]["id"] is always true

PR #512, opened 2025-07-15
Fixes for trailing entries and closes

TayyabTalha:


ozymotv:
This PR addresses an issue where init_pnls could fail if the account has no previous trade fills. It ensures that the bot initializes P&L correctly even when the fills list is empty, preventing crashes and enabling smoother startup in fresh or inactive accounts.

PR #513, opened 2025-07-18
WIP: Fix duplicate entries

Couillos:
The theory is calc_ideal_orders and execute_to_exchange functions are not detecting duplicated orders correctly because of their store signature
I used price step and qty step for rounding price and qty for the order signatures, and the Decimal module to avoid float rounding errors
There are some commented debug prints and boilerplate code, the purpose of this PR is to talk about this potential fix


Couillos:
Nevermind, this doesn't solve the issue

PR #529, opened 2025-11-14
add positions_held_per_day_w

backstab:
Add a new metric following the same _w pattern: positions_held_per_day_w


backstab:
Forgot to add it to optimize.py...

PR #533, opened 2025-11-27
Steps bounds

Couillos:
The idea of this feature is to explicitly specify steps for the parameters we want to optimize, in order to reduce the search space and overfitting. It may not make much sense to optimize some parameters like ema spans or log_range span at the minute precision.

What do you think of this current implementation? This PR is not meant to be merged as is, I would like your point of view

Thanks!

image


Couillos:
A more simple way of specifying steps could be:
```
"long_ema_span_0": [360, 1440, 30],
"long_ema_span_1": [360, 1440, 30],
```
What do you think?

enarjord:
I like it. Let each element in optimize.bounds be either a list of two floats or three. If two, current behavior. If three, last float is interpreted as step.

Couillos:
I prefer this notation too 👍. I'll rework it a little bit and test it further

PR #548, opened 2026-01-15
PR-547

FelixJongleur42:
cache folder name replacing ':' with '_' to ensure compatibility with Windows file naming restrictions

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,10 @@ Add your keys to api-keys.json.
 
 ### Step 7: Run Passivbot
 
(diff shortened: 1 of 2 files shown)


FelixJongleur42:
happy to get some feedback if such kind of PRs are desired.
As to my understanding, running under Windows would currently be broken without this change.
To make it clear: the env var WINDOWS_COMPATIBILITY is only necessary if you want to *force* this behavior.
If running directly under Windows, it automatically applies the folder name changes.

PR #562, opened 2026-02-15
fix: decouple warmup_minutes from HLCV cache hash for better reuse

enarjord:
## Summary

- Configs that differ only in trading parameters (EMA spans, warmup ratio) now share the same backtest HLCV cache slot, avoiding redundant re-downloads of candlestick data
- The cache uses a **ratchet-up strategy**: warmup sufficiency is checked at load time, and the cache is overwritten only when a larger warmup is needed
- Backward-compatible with existing caches (missing `cache_meta.json` treated as warmup=0)

## Problem

When backtesting two configs that share identical `backtest.*`, `live.approved_coins`, and exchange settings but differ in `bot.long` (e.g., different EMA spans), the HLCV data was re-downloaded from scratch. This happened because `warmup_minutes`—derived from the max EMA span—was included in the cache hash. Different EMA spans → different warmup → different hash → cache miss → full re-download.

Example with two real configs:
- **template_candidate.json**: `entry_volatility_ema_span_hours=672` → warmup=36,288 min (25.2 days)
- **template_current.json**: `entry_volatility_ema_span_hours=2490` → warmup=44,820 min (31.1 days)
- Same coins, same dates, same exchange → but produced different cache hashes

## Solution

1. **`get_cache_hash()`** — removed `warmup_minutes` from the hash. The key now depends only on data-defining parameters (coins, dates, exchange, coin age, gap tolerance, sources).

2. **`load_coins_hlcvs_from_cache()`** — new `warmup_minutes` parameter. Reads `cache_meta.json` and checks `cached_warmup >= needed_warmup`. Returns `None` if insufficient.

3. **`save_coins_hlcvs_to_cache()`** — new `warmup_minutes` parameter. Writes `cache_meta.json`. Removed the early-return guard so that re-fetches with larger warmup can overwrite the same cache slot.

4. **`prepare_hlcvs_mss()`** — threads `backtest_warmup_minutes` through to both load and save.

### Ratchet behavior

| Run | Config warmup | Cached warmup | Result |
|-----|--------------|---------------|--------|
| 1 | 36,288 | — | Miss → fetch → save (cached=36,288) |
| 2 | 36,288 | 36,288 | Hit |
| 3 | 44,820 | 36,288 | Miss (insufficient) → fetch → save (cached=44,820) |
| 4 | 36,288 | 44,820 | Hit (44,820 >= 36,288) |
| 5 | 44,820 | 44,820 | Hit |

## Test plan

- [x] 16 new tests in `tests/test_cache_warmup_reuse.py` covering:
  - Cache hash independence from EMA spans and warmup ratio
  - Non-warmup fields (coins, dates) still differentiate hashes
  - Warmup sufficiency gate (hit/miss/boundary)
  - Legacy cache compatibility (no `cache_meta.json`)
  - Corrupt `cache_meta.json` fallback
  - Save persists `cache_meta.json` (uncompressed + compressed)
  - Overwrite behavior (ratchet-up)
  - Full end-to-end ratchet scenario
- [x] Full test suite: 744 passed, 118 skipped, 0 failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable user-facing changes will be documented in this file.
 ### Added
 - **Hyperliquid builder codes** - Builder code attribution enabled by default (2 bps fee to fund development). Auto-approves with main wallet key; shows periodic nag banner for agent wallet users who haven't approved yet. Configurable via `broker_codes.hjson`.
 
+### Fixed
+- **Backtest HLCV cache reuse across configs** - Configs that differ only in trading parameters (EMA spans, warmup ratio) now share the same HLCV cache slot. Previously, different EMA spans produced different `warmup_minutes`, which was included in the cache hash, causing unnecessary re-downloads. The cache now uses a ratchet-up strategy: warmup sufficiency is checked at load time, and the cache is overwritten only when a larger warmup is needed.
+
 ## v7.8.2 - 2026-02-09
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+### Added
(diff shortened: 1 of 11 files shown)


enarjord:
Superseded by #564, which contains only the targeted HLCV cache warmup reuse changes on top of current master.

PR #573, opened 2026-03-21
Add entrypoint script for API key configuration and application execution

abendemar:


enarjord:
Thanks for this. The container/Railway deployment gap you were aiming to address has since been handled on `master`, but in a broader canonical way rather than as a Railway-specific runtime fork.

On April 4, 2026, `master` picked up `ac62b898` (`Implement canonical live container runtime contract`), which added the current `Dockerfile_live`, `container/entrypoint.sh`, env-generated `api-keys.json` support in `container/render_api_keys.py`, env-driven config rendering in `container/render_config.py`, the Compose example, Railway docs, and explicit container-runtime tests. That was then tightened by `e6f81af0` (`Fix container config path semantics and file logging`) and `6d5c0656` (`Use canonical logging config in container runtime`), and merged to `master` via `37776b11` on April 5, 2026.

The current approach also avoids the drift this PR was heading toward with a dedicated `railway/` config snapshot. Railway now reuses the same `passivbot live` container path documented in `docs/container_deployment.md` and `docs/hosting/railway.md`, and the design notes in `docs/plans/container_deployment_design.md` explicitly call out that a Railway-specific implementation had started drifting from the real config pipeline.

So I’m closing this PR as superseded by the recent container-runtime work on `master`. The suggestion was useful and helped surface the deployment need, but I don’t want to merge the older Railway-specific version now that the canonical implementation is in place.

PR #574, opened 2026-03-24
refactor: migrate optimizer from DEAP to pymoo NSGA-III

backstab:
- Replace DEAP evolutionary algorithm with pymoo's NSGA-III using Das-Dennis reference directions for better many-objective optimization
-  Break monolithic optimize.py into modular components: bounds.py, evaluator.py, problem.py, callback.py, output.py, repair.py
-  Replace discrete stepped bounds with continuous optimization + significant-digits rounding via SignificantDigitsRepair operator
-  Simplify ParetoStore from ~322 to ~45 lines (pymoo owns the Pareto front in-memory)

 Breaking changes

 - deap dependency replaced with pymoo>=0.6.0
 - Step values in [low, high, step] bounds are now ignored: optimization is continuous with rounding to N significant digits
 - Population size is adjusted to match NSGA-III reference direction requirements

 Test plan

 - New test suites for all extracted modules (bounds, callback, problem, repair, ref_dirs, pymoo integration)
 - Existing optimization tests updated


enarjord:
Thanks for this. The main direction in this PR has since been addressed on `master`, but not in exactly the same form.

`master` now has a first-class `pymoo` backend with NSGA-II / NSGA-III support, Das-Dennis reference directions, extracted optimizer modules under `src/optimization/`, and docs/config support for the nested `optimize.pymoo.*` settings. The main pieces landed across `e8d08b44` (`Wire pymoo nsga3 config and defaults`), `2511a740` (`Auto-size pymoo nsga3 population defaults`), `a8369543` (`Stream pymoo results as workers finish`), and `5915b7df` (`Make pymoo the default optimizer backend`).

The current code reflects that in `src/optimization/backends/pymoo_backend.py`, `src/optimization/backends/__init__.py`, and the modular `src/optimization/` package, and the user-facing behavior is documented in `docs/optimizing.md`.

The main reason I do not want to merge this PR now is that `master` intentionally kept compatibility where this PR proposed breaking changes. In particular, Passivbot still supports both optimizer backends (`deap` and `pymoo`), and stepped bounds are still preserved via `Bound.step` / quantization in `src/optimization/bounds.py` rather than being globally discarded. `requirements-full.txt` also still includes both `deap` and `pymoo`.

So I’m closing this PR as superseded by the optimizer work already merged to `master`. The proposal was useful and pointed in the right direction, but the current `master` implementation already covers the core pymoo/NSGA-III work while preserving the compatibility choices I want to keep.

PR #584, opened 2026-04-11
Fall back to ema_readiness for all-zero forager weights

backstab:
## Summary

`normalize_forager_score_weights` previously defaulted an all-zero weight vector to `{volume: 1.0, ema_readiness: 0.0, volatility: 0.0}`. Combined with `validate_forager_config` — which demands `forager_volume_ema_span > 0` whenever the normalized volume weight is positive — this turned "forager ranking disabled" into an impossible constraint for any user who locked both the score weights *and* the volume span to zero in `optimize.bounds`.

This pattern is common for single-coin setups where forager ranking is moot (there's nothing to rank, so users lock all the forager knobs to `[0, 0]`). The optimizer produces a pareto config with `forager_score_weights = {0, 0, 0}` and `forager_volume_ema_span = 0`, which then fails to load with:

```
ValueError: bot.long.forager_volume_ema_span must be > 0 when forager volume ranking or volume pruning is enabled
```

Fix: fall back to `{volume: 0.0, ema_readiness: 1.0, volatility: 0.0}` instead. `ema_readiness` is the only ranker that requires no auxiliary `forager_*_ema_span` — it's computed from the bot's regular entry EMAs (`ema_span_0` / `ema_span_1`) at `passivbot-rust/src/coin_selection.rs:140-178`. The log message in `normalize_bot_forager_config` that previously announced "volume-only" is updated to match.

## Reproduction

Any config where `optimize.bounds` locks all of:
- `long_forager_score_weights_{volume,ema_readiness,volatility}` to `[0, 0]`
- `long_forager_volume_ema_span` to `[0, 0]`

with `long_total_wallet_exposure_limit > 0` and `long_n_positions >= 1`.

## Test plan

- [x] Added `test_prepare_config_accepts_all_zero_long_forager_weights_when_long_enabled` — exercises the full `prepare_config` → `normalize_config` → `validate_forager_config` chain with all-zero long weights and pins the fallback output. Verified it reproduces the original ValueError without the fix applied.
- [x] `pytest tests/test_config_pipeline.py` — 26 passed
- [x] `pytest tests/ -k "forager or optimize or config"` — 343 passed

## Notes

- Behaviorally a no-op for existing valid configs: any config that previously relied on the `volume=1.0` fallback would have been tripping the validator already (the fallback and the validator were in direct conflict), so this fix only affects code paths that were previously unreachable.
- For multi-coin setups where the fallback matters semantically, `ema_readiness` is arguably a better default anyway — it biases empty slots toward coins closest to a real initial entry rather than toward the most-liquid tickers, which is usually what "I didn't specify a preference" should mean.

### DIFF
--- a/src/config/bot.py
+++ b/src/config/bot.py
@@ -177,7 +177,7 @@ def normalize_forager_score_weights(weights: dict, *, path: str) -> dict:
         total += value
 
     if total <= 0.0:
-        return {"volume": 1.0, "ema_readiness": 0.0, "volatility": 0.0}
+        return {"volume": 0.0, "ema_readiness": 1.0, "volatility": 0.0}
 
     return {key: normalized[key] / total for key in ("volume", "ema_readiness", "volatility")}
 
@@ -232,7 +232,7 @@ def normalize_bot_forager_config(
                 log_config_message(
                     verbose,
                     logging.INFO,
-                    "normalizing bot.%s.forager_score_weights all-zero vector to volume-only",
+                    "normalizing bot.%s.forager_score_weights all-zero vector to ema_readiness-only",
                     pside,
                 )
             else:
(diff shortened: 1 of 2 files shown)


enarjord:
Implemented and merged your idea into master: all-zero \ now fall back to EMA-readiness-only instead of arbitrary/legacy ranking. Merged in 0a966b34.

enarjord:
Implemented and merged your idea into master: all-zero forager_score_weights now fall back to EMA-readiness-only instead of arbitrary or legacy ranking. Merged in 0a966b34.

PR #587, opened 2026-04-11
refactor: consolidate warmup helpers in warmup_utils

backstab:
## Summary

- `src/downloader.py` carried a near-verbatim copy of `_to_float`, `_require_max_warmup_minutes`, `_iter_param_sets`, `compute_backtest_warmup_minutes`, and `compute_per_coin_warmup_minutes` that already existed in `src/warmup_utils.py`. The two copies had begun to drift — `warmup_utils._iter_param_sets` guards `bot_overrides.get("long", {}) or {}` against an explicit `None`, the downloader copy did not. Duplication was a bug magnet.
- Delete the downloader-local copies (and the now-orphaned `import math`) and migrate the 5 call-sites that had been importing from `downloader` — `src/optimize.py`, `src/optimize_suite.py`, `src/suite_runner.py`, `src/passivbot.py`, `tests/test_hlcvs_valid_ranges.py` — to import from `warmup_utils` directly. No shim or re-export; `warmup_utils` is now the single import path.
- **Latent effect worth noting:** callers previously routed through `downloader` silently inherit the `None`-guard in `_iter_param_sets`. They would have crashed on an explicit `coin_overrides[coin]["bot"]["long"] = None`; they now coerce to an empty dict like the `backtest` / `hlcv_preparation` path already did. This is an intrinsic consequence of picking `warmup_utils` as the survivor, not a deliberate behavioural change.

Net diff: 6 files changed, 5 insertions(+), 128 deletions(-).

## Test plan

- [x] `python -m pytest tests/test_warmup_utils.py tests/test_hlcvs_valid_ranges.py -x` — 27 passed
- [x] Import smoke test: `downloader`, `optimize`, `optimize_suite`, `suite_runner`, `passivbot`, `backtest`, `hlcv_preparation`, `warmup_utils` all import cleanly
- [x] `grep -rn "from downloader import compute_" src/ tests/` returns zero hits
- [x] Confirmed `downloader` module no longer exposes `math`, `compute_backtest_warmup_minutes`, or `compute_per_coin_warmup_minutes` as attributes

### DIFF
--- a/src/downloader.py
+++ b/src/downloader.py
@@ -5,7 +5,6 @@ import gzip
 import json
 import logging
 import inspect
-import math
 import os
 import shutil
 import sys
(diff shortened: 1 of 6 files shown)


enarjord:
Rebased this onto current master locally. After resolving the stale downloader.py conflict by keeping master's deletion, the branch collapsed to an empty diff against master, so this PR is moot now.

PR #594, opened 2026-05-29
refactor: replace deprecated logger.warn with logger.warning

MD-Mushfiqur123:
`logger.warn()` has been deprecated since Python 3.9 in favor of `logger.warning()`. This keeps the codebase modern and avoids DeprecationWarning noise.


enarjord:
Thanks for the cleanup PR. This appears to be moot against current `master`: the PR has no changed files, and `master` no longer contains any `logger.warn(...)` calls.

There were historical occurrences in the old manager code, but that code was removed in 2024. The remaining `.warn(` matches are `warnings.warn(...)`, which is the Python warnings API and should not be changed to `logger.warning(...)`.

Closing this as no-op.

PR #602, opened 2026-06-16
Feat: Add `--resume` flag to restore exact optimizer state

OzymoGit:
### Description
This PR introduces the ability to resume optimizations from the exact state where they left off, preventing the loss of progress due to sudden interruptions like power outages or crashes. 
Previously, if an optimization was stopped, all evolutionary progress was lost. Now, the optimizer continuously checkpoints its state and can be cleanly resumed.
### Key Changes
- **Checkpointing**: The optimizer now pickles its state (including `population`, `logbook`, `random` states, and PyMoo `Algorithm` state) to `checkpoint.pkl` inside the `results_dir` at the end of every generation.
- **CLI Flag**: Added a `--resume` argument to `optimize.py` which accepts the path to an existing `optimize_results/...` folder.
- **Backend Support**: Full resumption support for both the **DEAP** and **PyMoo** backends.
- **Result Appending**: Resumed optimizations will skip initial random evaluations, initialize from the checkpoint, and seamlessly append new records to the original `all_results.bin` file.
### How to use
To resume a previously interrupted optimization, run the original command and append `--resume` with the path to the interrupted results folder:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@ All notable user-facing changes will be documented in this file.
 - Fixed Hyperliquid `xyz:*` stock-perp backtest/optimizer startup so explicit
   `backtest.ohlcv_source_dir` data can use the direct source-dir preparation path when
   strict local v2 materialization is unavailable.
+- Added optimizer `--resume` checkpoint recovery safeguards: resume now requires a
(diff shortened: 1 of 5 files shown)


OzymoGit:
This is using a lot of memory. Could anyone please review it?

PR #605, opened 2026-06-24
Emit forager and EMA summary events

enarjord:
Observability-only logging slice from updated v8.

Scope:
- add structured event types for forager selection, forager feature unavailability, EMA bundle completion, EMA fallback use, and EMA unavailable summaries
- route those events to structured/monitor sinks only; console/text remain disabled by default
- emit bounded summaries from existing get_filtered_coins and orchestrator EMA bundle paths
- reuse already-computed counts/samples; no extra exchange calls and no trading-decision changes

Tests:
- python -m compileall src/live/event_bus.py src/live/event_emitters.py src/passivbot.py tests/test_live_event_bus.py tests/test_passivbot_monitor.py
- pytest tests/test_live_event_bus.py tests/test_passivbot_monitor.py tests/test_coin_filtering.py tests/test_live_candle_budget.py tests/test_missing_ema_fix.py -q
- git diff --check

Review focus:
- event emission must stay best-effort and behavior-neutral
- payloads must stay bounded and free of raw exchange/account payloads
- console/text defaults should remain quiet for EMA/forager internals

PR #597, opened 2026-06-04
Support Python 3.14 installs

vigilpbclaw-hash:
## Summary
- add Python 3.14-compatible dependency pins while preserving existing pins for older Python versions
- skip the old numba pin on Python 3.14, where it is not compatible and is not imported by the codebase
- build the existing PyO3 0.21 Rust extension with `abi3-py312` so Python 3.14 installs work without a PyO3 API migration

## Validation
- `python3.14 -m venv .venv314`
- `python -m pip install -e ".[dev]"`
- `python -m pip check`
- import smoke test for `numpy`, `pandas`, `matplotlib`, `psutil`, `passivbot_rust`
- `passivbot --help`
- `cd passivbot-rust && cargo check --tests`
- `python -m pytest -q tests/test_passivbot_cli.py tests/test_rust_utils.py tests/test_custom_endpoints.py`

## Notes
Full `pytest -q` under Python 3.14 still has unrelated failures in config/help-string tests and one monitor fake test. The install/build path succeeds.

PR #607, opened 2026-06-24
Detect abi3 Rust extension builds

enarjord:
Fix VPS startup after Python 3.14/abi3 Rust packaging change.\n\nThe Rust extension freshness guard only scanned the exact interpreter suffix when EXT_SUFFIX was present, so current maturin abi3 builds such as passivbot_rust.abi3.so were invisible. Live startup rebuilt successfully, then immediately failed because preferred_compiled_path() still returned None.\n\nChanges:\n- include abi3.so and generic shared-object fallbacks in extension suffix discovery\n- add a regression test matching the VPS package layout

### DIFF
--- a/src/rust_utils.py
+++ b/src/rust_utils.py
@@ -28,11 +28,13 @@ _LOCK_HANDLE = None
 
 def _extension_suffixes() -> list[str]:
     # Prefer the exact interpreter suffix when available (e.g. `.cpython-312-darwin.so`),
-    # but keep broad fallbacks for non-standard environments.
+    # but keep ABI-stable and broad fallbacks for non-standard environments.
     suffix = sysconfig.get_config_var("EXT_SUFFIX")
+    suffixes: list[str] = []
     if suffix:
-        return [suffix.lstrip(".")]
-    return ["so", "pyd", "dll", "dylib", "bundle", "sl"]
+        suffixes.append(suffix.lstrip("."))
+    suffixes.extend(["abi3.so", "so", "pyd", "dll", "dylib", "bundle", "sl"])
+    return list(dict.fromkeys(suffixes))
 
 
 def _local_extension_candidates() -> list[Path]:
--- a/tests/test_rust_utils.py
+++ b/tests/test_rust_utils.py
@@ -355,6 +355,37 @@ def test_preferred_compiled_path_uses_actual_import_target_for_package_layout(
     assert preferred_compiled_path() == installed
 
 
+def test_preferred_compiled_path_accepts_abi3_package_layout(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+):
+    site_packages = tmp_path / "site-packages"
+    package_dir = site_packages / "passivbot_rust"
+    package_dir.mkdir(parents=True)
+    (package_dir / "__init__.py").write_text("from .passivbot_rust import *\n")
+    installed = package_dir / "passivbot_rust.abi3.so"
+    installed.write_text("fresh-abi3-installed")
+
+    monkeypatch.chdir(tmp_path)
+    monkeypatch.setattr(
+        "rust_utils.sysconfig.get_config_var",
+        lambda name: ".cpython-312-x86_64-linux-gnu.so" if name == "EXT_SUFFIX" else None,
+    )
+    monkeypatch.setattr(
+        "rust_utils.sysconfig.get_paths",
+        lambda: {"platlib": str(site_packages), "purelib": str(site_packages)},
+    )
+    monkeypatch.setattr(
+        "rust_utils.importlib.util.find_spec",
+        lambda name: SimpleNamespace(
+            origin=str(package_dir / "__init__.py"),
+            submodule_search_locations=[str(package_dir)],
+        ),
+    )
+
+    assert _installed_extension_candidates() == [installed]
+    assert preferred_compiled_path() == installed
+
+
 def test_prune_shadowing_local_extensions_removes_src_copy_when_installed_exists(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ):


enarjord:
Reviewer: Claude (automated PR reviewer)
PR: #607
Head SHA: aadf8ff6c89d1f86bff1bffb289bf0017155d33e
Base SHA/branch: 30a07c9908d8c7baaf805d6dc1c8f500dae22f16 (v8; incl. merged #605)
Scope reviewed: full PR (first review) — single commit, src/rust_utils.py (+8/-3) + tests/test_rust_utils.py (+31)
Verdict: approve

Correct, well-targeted, and well-tested. **Build/tooling only — no live trading, order, risk, or backtest logic.** This is the natural follow-up to #597: now that the Rust ext is built with `abi3-py312`, the installed artifact is `passivbot_rust.abi3.so`, not the interpreter-specific `passivbot_rust.cpython-3XX-<plat>.so`.

### Why it's correct
- Old `_extension_suffixes()` returned **only** the exact `EXT_SUFFIX` when set (the normal CPython case), so an abi3 artifact would never be globbed by `_local/_installed/_target_extension_candidates()` → the extension would be "missed" (staleness checks, prune, mtime, preferred-path resolution all silently fail to find it). The fix always appends `["abi3.so", "so", "pyd", ...]` and dedups the suffix list with `dict.fromkeys`.
- `abi3.so` is ordered before `so`, so a `*.abi3.so` file is matched by the more specific suffix first. `dict.fromkeys` also correctly collapses the case where `EXT_SUFFIX` itself is `.abi3.so` (limited-API builds) so it isn't listed twice.

### Duplicate-path side effect — checked, benign
Because the broad fallbacks are now always included, a single local file can match two suffix globs (e.g. `passivbot_rust.cpython-312-...so` matches both the exact-suffix glob and `so`), and `_local_extension_candidates()` is the one finder that does not dedup. I verified every consumer tolerates duplicates:
- `prune_shadowing_local_extensions` guards each unlink with `if local.exists()` + `unlink(missing_ok=True)` + `except OSError` — a repeated path is a no-op the second time.
- `preferred_compiled_mtime`/`preferred_compiled_path` use `max(...)` over the group — duplicates don't change the result.
- `_installed_extension_candidates` already dedups via `dict.fromkeys`.

### Tests I ran (worktree at aadf8ff6, repo venv)
- `tests/test_rust_utils.py`: **21 passed**, incl. the new `test_preferred_compiled_path_accepts_abi3_package_layout` (sets `EXT_SUFFIX` to a non-matching cpython suffix and asserts the `passivbot_rust.abi3.so` package-layout file is detected and preferred).

### Findings
- Blocker: none.
- Important: none.
- Minor (optional): consider deduping `_local_extension_candidates()` with `list(dict.fromkeys(out))` for parity with `_installed_extension_candidates()` — purely cosmetic now (all consumers are duplicate-safe), but it keeps `compiled_extension_paths()` free of repeats.

vigilpbclaw-hash:
PR: #607
Head SHA: aadf8ff6c89d1f86bff1bffb289bf0017155d33e
Base SHA/branch: 81d4ceb19b72d6d3ef3db4ed0dfe078eab59a6f4/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Evidence:
- Reviewed `src/rust_utils.py::_extension_suffixes()` and the new regression in `tests/test_rust_utils.py::test_preferred_compiled_path_accepts_abi3_package_layout`.
- The change is build/startup tooling only; I found no trading behavior, risk/order logic, event-pipeline, or data freshness/coverage impact.
- Keeping the exact interpreter suffix first while appending `abi3.so` and generic shared-object fallbacks preserves existing preference and fixes detection for maturin abi3 package-layout builds. Deduplication prevents repeated suffix scans.
- The test models the failing case where `EXT_SUFFIX` is interpreter-specific but the installed artifact is `passivbot_rust.abi3.so`, and asserts both candidate discovery and `preferred_compiled_path()` select it.

Checks run:
- `git diff --check 81d4ceb19b72d6d3ef3db4ed0dfe078eab59a6f4...aadf8ff6c89d1f86bff1bffb289bf0017155d33e` — passed.
- `python3 -m pytest tests/test_rust_utils.py -q` — 21 passed, 1 pre-existing config warning about unknown `asyncio_mode` under the local pytest/Python environment.

enarjord:
Reviewer: Cursor Agent (continuous reviewer)
PR: #607
Head SHA: aadf8ff6c89d1f86bff1bffb289bf0017155d33e
Base SHA/branch: 81d4ceb19b72d6d3ef3db4ed0dfe078eab59a6f4 / v8
Scope reviewed: full PR
Verdict: approve

## Findings

### Blocker
- None.

### Important
- None. Fix is narrowly scoped to extension discovery for abi3 maturin layouts; exact `EXT_SUFFIX` still preferred first.

### Minor
- `abi3.so` is now always in the fallback list even when `EXT_SUFFIX` is set — harmless dedupe via `dict.fromkeys`, and required for `passivbot_rust.abi3.so` package layouts.

## Review notes
- **Trading/risk impact:** None (startup/packaging path in `rust_utils.py`).
- **Regression test:** `test_preferred_compiled_path_accepts_abi3_package_layout` matches VPS package layout.

## Checks run (PR head `aadf8ff6`, `/tmp/passivbot-pr606` venv)
- `pytest tests/test_rust_utils.py::test_preferred_compiled_path_accepts_abi3_package_layout tests/test_rust_utils.py::test_preferred_compiled_path_uses_actual_import_target_for_package_layout -q` — 2 passed

enarjord:
Closing as superseded by origin/v8 commit 81d4ceb1 (Fix abi3 Rust extension freshness checks), which implements the same startup fix with broader importlib suffix coverage and tests. VPS smoke on this branch confirmed the root cause; restarting VPS on current v8 next.

PR #612, opened 2026-06-24
Fix Bybit entry-grid price-limit rejects

msei99:
## Summary
- Add a Bybit-specific reconciler hook to skip live entry-grid limit buy creates below Bybit's dynamic lower price band before submission.
- Keep cancels untouched and fail loudly if required Bybit risk or live-price inputs are missing or invalid.
- Log skipped creates with throttled INFO summaries and DEBUG details.

## Tests
- `/home/mani/software/venv_pb7/bin/python -m pytest tests/exchanges/test_ccxt_bot_hooks.py::TestBybitCreateCcxtSessions::test_price_limit_guard_skips_only_too_deep_entry_grid_buys`
- `/home/mani/software/venv_pb7/bin/python -m pytest tests/exchanges/test_ccxt_bot_hooks.py`

## Live verification
- Smoke-tested on `manibot56` with `bybit_CPTV7HR`; POPCAT invalid creates were skipped by the guard and no new `110003` rejects appeared during the checked window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Fixed Bybit live entry-grid creation so buy orders below Bybit's dynamic
+  lower price band are skipped before submission and logged with throttled
+  skipped counts plus planned prices, avoiding repeated `110003` invalid-price
+  rejects.
 - Fixed Bitget UTA / Elite close-order placement by omitting the one-way-only
   `reduceOnly` flag from hedge-mode v3 orders that already send `posSide`.
 - Fixed Bitget UTA / Elite open-order normalization so hedge-mode close orders
--- a/src/exchanges/bybit.py
+++ b/src/exchanges/bybit.py
@@ -3,6 +3,7 @@ from passivbot import logging, clip_by_timestamp
 from uuid import uuid4
 import asyncio
 import ccxt
+import math
 from copy import deepcopy
 from collections import defaultdict
 from utils import symbol_to_coin, ts_to_date, utc_ms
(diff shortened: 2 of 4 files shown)


enarjord:
Reviewer: Cursor Agent (continuous reviewer)
PR: #612
Head SHA: 2f8246eea7240e64db10d8227833fdb5ea37c319
Base SHA/branch: b5c2e15bc0d649bedd54c28d7f38057bc403f734 / master
Scope reviewed: full PR
Verdict: approve

## Findings

### Blocker
- None.

### Important
- **Scope is intentionally narrow:** only non-reduce `entry_grid_*` **buy** creates are guarded. Other buy types or sell-side entry-grid creates outside Bybit's upper band are unchanged — acceptable if live issue was only lower-band `110003` rejects, but worth tracking if sell-side rejects appear.
- **Trading impact:** invalid deep entry-grid buys are dropped pre-submit instead of exchanged-rejected; cancels untouched. Missing `priceLimitRatioX`/live price/ratio inputs **fail loudly** (`ValueError`) per error contract.

### Minor
- Reconciler summary now uses `len(to_cancel)` and adds `exchange_price_limit_skipped` to skipped totals — accounting fix aligns totals with post-hook lists.
- INFO summaries throttled to 1h per `(coin, pside, type, prices)` key; DEBUG on repeat.

## Review notes
- Hook wired generically via `getattr(bot, "_filter_exchange_price_limit_creations")` — Bybit-only implementation.
- No silent fallbacks on guard inputs; skip path is explicit for known-invalid prices only.

## Checks run (PR head `2f8246ee`, `/tmp/passivbot-pr606` venv)
- `pytest tests/exchanges/test_ccxt_bot_hooks.py::TestBybitCreateCcxtSessions::test_price_limit_guard_skips_only_too_deep_entry_grid_buys` — passed
- `pytest tests/exchanges/test_ccxt_bot_hooks.py -q` — 59 passed

vigilpbclaw-hash:
PR: #612
Head SHA: 2f8246eea7240e64db10d8227833fdb5ea37c319
Base SHA/branch: b5c2e15bc0d649bedd54c28d7f38057bc403f734/master
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: None.
- Important: None.
- Minor: `src/exchanges/bybit.py::_filter_exchange_price_limit_creations` intentionally guards only non-reduce `entry_grid_*` buy creates below Bybit's lower band. That matches the PR scope and live `110003` symptom; if Bybit starts rejecting short/sell entry-grid creates above the upper band, this hook will need a symmetric follow-up.

Review notes:
- The exchange hook is called after existing order-match/initial-entry guards and before sorting/freshness checks, returns filtered creates plus an explicit skipped count, and leaves cancels untouched.
- Missing or invalid Bybit risk parameters, live price, symbol, or order price raise `ValueError`; I did not find a silent fallback that would hide data-freshness/risk input issues.
- Log payloads contain symbol/pside/type/prices/market/min_buy/ratio only; no raw exchange payloads or credentials. INFO summaries are throttled, repeated messages downgrade to DEBUG.
- Current base merge dry-run was clean.

Checks run:
- `git diff --check origin/master...HEAD` — passed.
- `git merge --no-commit --no-ff refs/remotes/origin/pr-612` from a fresh `/tmp` worktree at current `origin/master` — clean.
- `python3 -m pytest tests/exchanges/test_ccxt_bot_hooks.py::TestBybitCreateCcxtSessions::test_price_limit_guard_skips_only_too_deep_entry_grid_buys -q` — passed.
- `python3 -m pytest tests/exchanges/test_ccxt_bot_hooks.py -q` — 58 passed.

enarjord:
Reviewer: Claude (automated PR reviewer)
PR: #612
Head SHA: 2f8246eea7240e64db10d8227833fdb5ea37c319
Base SHA/branch: b5c2e15bc0d649bedd54c28d7f38057bc403f734 (master)
Scope reviewed: full PR (first review) — CHANGELOG (+4), src/exchanges/bybit.py (+125), src/live/reconciler.py (+11/-1), tests/exchanges/test_ccxt_bot_hooks.py (+59)
Verdict: needs fix

This is a **trading/order-logic change** (it skips entry-grid creates before submission), so I reviewed the order-path safety closely. The core idea is sound and conservative — Bybit-only (opt-in via `getattr` hook, other exchanges unaffected), narrowly scoped to non-reduceOnly entry-grid limit **buys**, skip-only (never mutates/adds orders, builds a new list via `id()`), logged rather than silent, and the skips are counted into the plan summary. But the error handling makes it unsafe in the live reconcile hot path.

### Important 1 — fail-loud guard aborts the entire cancel+create cycle
`_filter_exchange_price_limit_creations` raises `ValueError` on several data conditions: missing `symbol`, missing `markets_dict[...]["info"]["riskParameters"]["priceLimitRatioX"]`, missing live price for a symbol, missing order `price`, or `ratio` not in `(0,1)`. That hook is awaited **unwrapped** in `calc_orders_to_cancel_and_create` (reconciler.py:813-820; the function body, def at :783, has no enclosing try), which is itself called unwrapped in `execute_to_exchange` (executor.py:37) **before any orders are placed/cancelled**. So a single problematic symbol aborts the whole cycle for **all** symbols — no cancels and no creates that iteration. If the condition persists (e.g. one Bybit symbol whose `riskParameters` lacks `priceLimitRatioX`, a momentarily-missing live price, or a contract whose ratio isn't `<1`), order management halts every cycle.

Reproduced on this head (worktree at 2f8246ee):
```
markets_dict[sym]["info"]["riskParameters"] = {}   # priceLimitRatioX missing
_filter_exchange_price_limit_creations([], [entry_grid_buy])
  -> RAISED ValueError: X/USDT:USDT: missing Bybit priceLimitRatioX for price-limit guard
```
Fix: fail **safe** per order — if any input for an order can't be resolved (missing risk param / price / out-of-range ratio), skip *guarding that order* (let it through to the exchange, i.e. preserve today's behavior and let Bybit reject the rare bad one) and log it, instead of raising. A guard that prevents `110003` rejects should never be able to stop the bot from managing orders entirely.

### Important 2 — throttle key includes the exact prices, so it neither throttles nor bounds memory
`log_key = (coin, pside, pb_type, prices)` where `prices` is the comma-joined skipped order prices. Entry-grid prices move with the market/position, so the key changes on most cycles → `last_info_ms` is 0 → the `>= 60*60*1000` check passes → it logs at **INFO every cycle**, defeating the "throttled" intent (CHANGELOG says "throttled skipped counts") and producing INFO spam. It also means `self._bybit_price_limit_guard_info_ms` gains a new entry per distinct price set and is never evicted → unbounded growth over a long-running live bot. Fix: key the throttle on `(coin, pside, pb_type)` (drop `prices`), and/or bound/evict the dict.

### Minor 3 — only the long side / lower band is handled
The guard covers entry-grid **buys** below the lower band. Short entry-grid **sell** creates above Bybit's upper price band can still hit `110003`; they aren't filtered. Not a regression (this fixes the long side), but the symmetric upper-band/sell case remains a follow-up.

### Minor 4 — correctness hinges on `priceLimitRatioX` semantics
The math assumes `priceLimitRatioX` is the lower-bound fraction, so `min_buy = last_price * ratio` (test models `0.1` → band at 10% of last). Since the guard only skips, the risky direction is **over-skipping** legitimate deep-grid buys if the field ever means something other than a `(0,1)` lower-bound fraction. Combined with Important 1 (fail-loud on `ratio >= 1`), I'd want this validated against live Bybit `riskParameters` across contract types before relying on it. Worth a short code comment documenting the expected field semantics + units.

### Tests I ran (worktree at 2f8246ee, repo venv)
- `tests/exchanges/test_ccxt_bot_hooks.py -k price_limit_guard`: 1 passed. Good happy-path coverage (skips the too-deep buy, keeps the in-band one, leaves cancels untouched, asserts one INFO line). No test exercises the raise paths or the missing-data degradation — adding those would lock in the fail-safe behavior from Important 1.

enarjord:
Thanks for the fix and for identifying the Bybit `110003` failure mode.

I opened #620 as a master-targeted replacement with a universal live guard instead of a Bybit-only guard:

- new config knob: `live.limit_order_create_max_market_dist_pct`
- default `0.8`, so live skips limit creates below `market * 0.2` for buys or above `market * 1.8` for sells
- applies across exchanges and order types before create submission, using a fresh market snapshot
- keeps market orders and stale-snapshot failure handling unchanged

That should cover the reported deep-grid spam without hardcoding Bybit-specific behavior in this first pass. If we later want exact venue limits, I think the cleaner next step is a separate dynamic exchange-band layer on top of this general fallback, using Bybit `riskParameters.priceLimitRatioX` and equivalent fields where available.

Closing this as superseded by #620.

PR #722, opened 2026-06-27
Add cache doctor candle coverage evidence

enarjord:
## Summary
- add read-only v2 candle coverage windows to cache-integrity-doctor from local .valid.npy masks
- include bounded suspicious interior gap samples and warm-cache evidence labels in candle family summaries
- update cache-doctor docs/backlog/changelog for the first coverage slice

## Validation
- pytest tests/test_cache_integrity_doctor.py
- python -m compileall src/tools/cache_integrity_doctor.py tests/test_cache_integrity_doctor.py
- git diff --check
- rg -n "except Exception|return_exceptions=True|\.get\([^\n]*,\s*(0|0\.0|None|False|\{\}|\[\])\)" src/tools/cache_integrity_doctor.py tests/test_cache_integrity_doctor.py (no matches)

## Boundaries
- local/read-only cache artifact inspection only
- no live bot control, exchange access, credentials, SSH, or trading behavior changes

PR #720, opened 2026-06-27
Add EMA live event debug profile

enarjord:
## Summary
- add `ema` as a live-event debug profile, including `ema-readiness` aliases
- enrich `ema.unavailable` structured events with bounded parsed EMA type/span/inner-reason detail when `ema` or `candles` debug profiles are enabled
- keep default `ema.unavailable` payloads compact and keep console/trading behavior unchanged

## Safety / Scope
- observability-only; no exchange calls, no order/risk behavior changes
- debug enrichment uses data already captured for existing `ema.unavailable` events
- no new console output and no raw payload persistence

## Validation
- `./venv/bin/python -m pytest tests/test_live_event_bus.py tests/test_passivbot_monitor.py` -> 86 passed, 1 warning
- `./venv/bin/python -m compileall -q src/live/event_bus.py src/live/event_emitters.py tests/test_live_event_bus.py tests/test_passivbot_monitor.py` -> pass
- `git diff --check` -> pass
- touched-file silent-handling scan reports existing best-effort event emitter/sink patterns only; no new trading-path fallback

PR #719, opened 2026-06-27
Add live restart smoke plan tool

enarjord:
## Summary
- add `passivbot tool live-restart-smoke-plan`, a read-only dry-run planner for the repeated live restart/smoke routine
- expose sanitized tmuxp live command parsing for reuse by the planner
- document the first restart-smoke automation slice in the backlog and changelog

## Safety / Scope
- plan-only: no SSH, no tmux invocation, no process signaling, no git pull/fetch/checkout, no `passivbot live` startup, no exchange/API calls, no credential loading
- `--execute` is explicitly rejected
- no trading behavior changes

## Validation
- `./venv/bin/python -m pytest tests/test_live_restart_smoke_plan.py tests/test_live_smoke_report.py tests/test_passivbot_cli.py` -> 85 passed, 13 warnings
- `./venv/bin/python -m compileall -q src/live/restart_smoke_plan.py src/tools/live_restart_smoke_plan.py src/live/smoke_report.py src/passivbot_cli/main.py` -> pass
- `git diff --check v8..HEAD` -> pass

PR #718, opened 2026-06-26
Add live event debug profiles

enarjord:
## Summary
- add `logging.live_event_debug_profiles` and `PASSIVBOT_LIVE_EVENT_DEBUG_PROFILES`
- add the first `rust` profile, enriching structured Rust orchestrator events with bounded input-symbol and output-order samples
- keep default behavior compact: no console routing changes, no raw Rust payload persistence, no trading behavior changes

## Validation
- `python -m compileall src/live/event_bus.py src/live/event_emitters.py src/passivbot.py src/config/schema.py`
- `PYTHONPATH=src ./venv/bin/python -m pytest tests/test_live_event_bus.py tests/test_passivbot_monitor.py -q`
- `git diff --check v8..HEAD`
- config normalization smoke for logging.live_event_debug_profiles=rust,remote-calls

## Notes
- `black` and `ruff` are not installed in this venv; formatting was checked manually plus `git diff --check`.
- This is observability-only. The profile only changes structured event data when explicitly enabled.

PR #723, opened 2026-06-27
Add remote-call live event debug profile

enarjord:
## Summary
- add opt-in `remote_calls` live-event debug profile enrichment for candle and authoritative remote-call events
- include bounded payload-shape/correlation details only: data keys, param keys, stage/surface/kind/timing IDs/counts
- keep default events unchanged and avoid raw payload values or console output

## Validation
- `./venv/bin/python -m pytest -q tests/test_passivbot_monitor.py::test_candle_remote_fetch_callback_emits_correlated_remote_call_events tests/test_passivbot_monitor.py::test_remote_call_debug_profile_adds_bounded_candle_payload_shape tests/test_passivbot_monitor.py::test_authoritative_timed_fetch_emits_correlated_remote_call_events tests/test_passivbot_monitor.py::test_remote_call_debug_profile_adds_authoritative_payload_shape`
- `./venv/bin/python -m pytest -q tests/test_live_event_bus.py tests/test_passivbot_monitor.py`
- `./venv/bin/python -m compileall -q src/live/event_emitters.py tests/test_passivbot_monitor.py`
- `git diff --check`
- touched-file silent-handling audit reviewed; matches are pre-existing best-effort event/test patterns, no trading fallback added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ All notable user-facing changes will be documented in this file.
 - Added the `ema` live-event debug profile, which enriches structured
   `ema.unavailable` events with bounded parsed readiness detail without
   changing console output or trading behavior.
+- Added the `remote_calls` live-event debug profile, which enriches structured
+  remote-call events with bounded payload-shape and correlation details without
+  adding raw payloads or console output.
 - Added `passivbot tool live-config-preflight` for read-only offline summaries
   of risk-relevant live config facts before startup.
 - Added shutdown-event summaries to `passivbot tool live-smoke-report`, including
--- a/docs/plans/live_ops_improvement_backlog.md
+++ b/docs/plans/live_ops_improvement_backlog.md
@@ -255,8 +255,8 @@ Related detailed plans:
     surfaces are touched.
 
 12. [ ] Debug profile toggles.
-    Status: partial. Initial Rust live-event debug profile slice merged; EMA
-    readiness profile slice is in progress.
+    Status: partial. Rust, EMA readiness, and remote-call profile slices are
+    merged or in progress.
 
     Add narrow runtime/debug profiles that increase event detail for one domain:
     candles, fills, HSL, Rust payloads, order execution, or exchange calls. This
@@ -272,10 +272,14 @@ Related detailed plans:
       `ema.unavailable` structured events with bounded parsed EMA type, span,
       and inner reason summaries while keeping default events compact and
       console output unchanged.
+    - 2026-06-26: Added `remote_calls` debug-profile enrichment for candle and
+      authoritative remote-call events, exposing bounded payload key shape,
+      parameter key names, and correlation state without raw payloads or
+      console output.
 
-    Remaining refinements: add targeted profiles for remote calls, candle
-    coverage/tail state, HSL, fills, and execution as those diagnostics need
-    deeper live evidence.
+    Remaining refinements: add targeted profiles for candle coverage/tail
+    state, HSL, fills, and execution as those diagnostics need deeper live
+    evidence.
 
 13. [ ] Cache integrity doctor.
     Status: partial. Initial read-only local cache smoke doctor and
--- a/src/live/event_emitters.py
+++ b/src/live/event_emitters.py
@@ -104,6 +104,49 @@ def _remote_fetch_payload_key(payload: dict[str, Any]) -> tuple[Any, ...]:
     )
 
 
+def _mapping_key_sample(value: Any, *, limit: int = 32) -> list[str]:
+    if not isinstance(value, dict):
+        return []
+    return sorted(str(key) for key in value)[:limit]
+
+
+def _remote_call_debug_payload(
+    data: dict[str, Any],
+    *,
+    matched_start: bool | None = None,
+    limit: int = 32,
+) -> dict[str, Any]:
+    debug: dict[str, Any] = {
+        "data_keys": _mapping_key_sample(data, limit=limit),
+    }
+    if isinstance(data.get("params"), dict):
+        debug["param_keys"] = _mapping_key_sample(data.get("params"), limit=limit)
+    if matched_start is not None:
+        debug["matched_start"] = bool(matched_start)
+    for key in (
+        "kind",
+        "surface",
+        "stage",
+        "tf",
+        "timeframe",
+        "since_ts",
+        "end_ts",
+        "limit",
+        "rows",
+        "first_ts",
+        "last_ts",
+        "elapsed_ms",
+        "state_epoch",
+        "url_hash",
+        "error_type",
+    ):
+        if key in data:
+            debug[key] = data.get(key)
+    if isinstance(data.get("pending_confirmations"), list):
+        debug["pending_confirmation_count"] = len(data["pending_confirmations"])
+    return debug
+
+
 def _remote_call_map_max(bot: Any) -> int:
     try:
         raw = getattr(bot, "_live_event_remote_call_map_max", None)
@@ -198,6 +241,12 @@ def emit_candle_remote_fetch_event(bot: Any, payload: dict[str, Any]) -> Any:
     data = _sanitize_remote_fetch_payload(payload)
     if stage != "start" and not matched_start:
         data["orphan_result"] = True
+    if live_event_debug_profile_enabled(bot, "remote_calls"):
+        data["debug_profile"] = "remote_calls"
+        data["debug"] = _remote_call_debug_payload(
+            data,
+            matched_start=matched_start if stage != "start" else None,
+        )
     return bot._emit_live_event(
         event_type,
         level=level,
@@ -293,6 +342,9 @@ def _emit_authoritative_remote_call_event_unchecked(
         data["error_repr"] = _sanitize_remote_text(repr(error), max_len=500)
     elif stage != "start":
         data.update(_authoritative_result_summary(surface, result))
+    if live_event_debug_profile_enabled(bot, "remote_calls"):
+        data["debug_profile"] = "remote_calls"
+        data["debug"] = _remote_call_debug_payload(data)
     emit(
         event_type,
         level=level,
--- a/tests/test_passivbot_monitor.py
+++ b/tests/test_passivbot_monitor.py
@@ -1790,7 +1790,13 @@ def test_forager_and_ema_summary_emitters_are_best_effort_on_malformed_inputs(ca
     assert any(EventTypes.BOT_STARTUP_TIMING in msg for msg in messages)
 
 
-def _make_remote_fetch_event_bot(sink, *, cycle_id="cy_7", map_max=None):
+def _make_remote_fetch_event_bot(
+    sink,
+    *,
+    cycle_id="cy_7",
+    map_max=None,
+    debug_profiles=(),
+):
     import passivbot as pb_mod
 
     class FakeBot:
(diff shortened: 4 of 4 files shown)


vigilpbclaw-hash:
PR: #723
Head SHA: e78d79b04fd36fb5a0dae2e15bb8d966a1504374
Base SHA/branch: cab54911106f40b27da25d7bfd7b48e8ab91ca4d/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: This PR appears to add opt-in remote-call debug enrichment for candle remote fetch and authoritative state fetch events while preserving default event payloads. The implementation fits the existing live event debug-profile abstraction (`live_event_debug_profile_enabled`) and keeps enrichment localized to event emitters rather than trading/order paths. The added data is bounded to keys/counts/selected timing/correlation fields, with existing URL/error sanitization preserved, so it remains observability-only.

Validation run:
- `python3 -m pytest -q tests/test_passivbot_monitor.py::test_candle_remote_fetch_callback_emits_correlated_remote_call_events tests/test_passivbot_monitor.py::test_remote_call_debug_profile_adds_bounded_candle_payload_shape tests/test_passivbot_monitor.py::test_authoritative_timed_fetch_emits_correlated_remote_call_events tests/test_passivbot_monitor.py::test_remote_call_debug_profile_adds_authoritative_payload_shape` — passed (4 tests).
- `python3 -m compileall -q src/live/event_emitters.py tests/test_passivbot_monitor.py` — passed.

enarjord (commented):
**Reviewer:** Claude reviewer loop
**PR:** #723 — Add remote-call live event debug profile
**Head SHA:** e78d79b04fd36fb5a0dae2e15bb8d966a1504374
**Base:** v8 (branch base `cab54911`, slightly behind v8 tip `3b7e6306`; no overlap with intervening merges — path-scoped review of the single PR commit)
**Scope reviewed:** `src/live/event_emitters.py` (+52), `tests/test_passivbot_monitor.py` (+103/-1), `CHANGELOG.md`, `docs/plans/live_ops_improvement_backlog.md`
**Verdict:** approve

Opt-in `remote_calls` debug-profile enrichment for candle and authoritative remote-call events. Clean, well-tested, and verified observability-only. Mirrors the existing `ema` profile slice (same `live_event_debug_profile_enabled` gate, same bounded-enrichment shape) — consistent and the right approach.

**Observability-only — verified (both paths exception-isolated):**
- Candle (`emit_candle_remote_fetch_event`): enrichment only adds `debug_profile`/`debug` keys to the `data` dict already passed to `_emit_live_event`. The callback is invoked via `CandlestickManager._emit_remote_fetch`, which wraps `cb(payload)` in `try/except Exception: return` (\"Must never break the fetch path\") — any failure in the debug build cannot touch the fetch path.
- Authoritative (`_emit_authoritative_remote_call_event_unchecked`): enrichment sits inside the `_unchecked` body, fully wrapped by `emit_authoritative_remote_call_event` (try/except → `logging.debug`, returns `remote_call_id`). Original fetch/exception and the return value are unaffected.
- No control-flow or return-value change on either path; no level/route change (warning/debug preserved) → no new console output.

**Opt-in / default-off:** gated on `live_event_debug_profile_enabled(bot, \"remote_calls\")`; `remote_calls` is registered in `LIVE_EVENT_DEBUG_PROFILES` (+ `remote`/`remote_call`/`remote-call(s)` aliases). Test confirms profile-off ⇒ no `debug` key on either event.

**Secret-safe — no new leakage surface:** the `debug` sub-dict is a bounded shape-summary re-derived from the already-emitted `data` (itself already sanitized for `error`/`error_repr`/`url`). It surfaces only: key *names* (`data_keys`, `param_keys`; sorted, capped at 32), a metadata/timing/identifier whitelist (`kind/surface/stage/tf/timeframe/since_ts/end_ts/limit/rows/first_ts/last_ts/elapsed_ms/state_epoch/url_hash/error_type`), `matched_start` (bool), and `pending_confirmation_count` (len). No raw param *values* — the new test asserts the param value `secret` never appears while only the key name `apiKey` is surfaced. Everything it copies is a subset of what `data` already emits.
- Note: `matched_start if stage != \"start\" else None` short-circuits correctly, so no NameError on the start branch (where `matched_start` is unbound).

**No registry drift:** reuses existing `REMOTE_CALL_*` event types + `REMOTE_CALL`/`CANDLE` tags; no new EventTypes/Tags/ReasonCodes.

**Docs:** CHANGELOG + backlog item-12 update are accurate and honest (records the slice, drops remote-calls from remaining-refinements).

**Tests run** (worktree @ `e78d79b0`, `--rootdir` = worktree): `tests/test_live_event_bus.py` + `tests/test_passivbot_monitor.py` → **88 passed** (includes the 4 PR-cited tests plus the two new debug-profile tests: bounded candle/authoritative shape + secret-non-leak, and the negative profile-off assertions added to the existing correlation tests).

Minor (observation, non-blocking): `data[\"debug_profile\"] = \"remote_calls\"` is set *before* `_remote_call_debug_payload(data)` runs, so `data_keys` self-includes the literal `\"debug_profile\"` marker (but not `\"debug\"`). Harmless cosmetic noise in the opt-in payload; if you want `data_keys` to reflect the true payload shape, compute `debug` before setting the marker, or exclude the marker keys in `_mapping_key_sample`. Not worth a respin.

Reviewed at: 2026-06-27T01:08:42Z

enarjord:
**Retrospective audit — issues in PRs #713–#722 (merged while the reviewer loop was offline)**

_Posted on #723 only because it is the sole open PR, so the implementor loop picks it up — this is **not** a re-review of #723 (that was approved separately above). All findings below were verified by hand against merged code at current v8 tip `3b7e6306`. Severity = follow-up priority on already-merged code. Docs-only #713/#716/#721 reviewed: accurate, no overclaim._

Two cross-cutting themes plus one standalone correctness gap.

---

### Theme A — path/username hygiene in shareable operator output

New operator tools each roll their own path rendering and surface absolute deploy paths / usernames / config filenames into output the PRs describe as shareable, instead of the existing `_user_safe_display_path` (collapses `$HOME` / `/root` / `/home/<u>` / `/Users/<u>` → `~/…`).

- **Important — #719** `src/live/restart_smoke_plan.py:23-26`: `_display_path()` = `str(Path(value).expanduser())` actively **expands** `~`→absolute (the opposite of redaction), applied to `inputs.{supervisor_config,repo_path,monitor_root,logs_root}` at `:248-251`, which land in the shippable plan JSON. Passing `~/passivbot` yields `/Users/<user>/passivbot`; passing `/root/bots_vps5.yaml` is surfaced verbatim where the convention would render `~/bots_vps5.yaml`. This module doesn't import `_user_safe_display_path` at all. Fix: route the four `inputs.*` path fields through `_user_safe_display_path` (raw paths inside the emitted copy-paste shell commands are fine — those need to run).
- **Important — #715** `src/live/smoke_report.py:1111-1115`: the new `latest_message` echoes the event's free-form `message` via `_redact_log_text` (credential-only, no home-collapse). It is the **only** report section that echoes raw event message text — peer sections `risk_events` / `problem_events` deliberately omit it. A shutdown `message` containing a deploy path / username / account name ships into `smoke_report.json`. (Note: the same file defines and uses `_user_safe_display_path` at `:139`/`:1407`. The sibling `data` string fields `reason`/`stage`/`error` at `:1058-1086` also only `_redact_log_text`, but that *matches* how existing `config_path`/`account` fields are already handled at `:1253-1261`, so that part is convention-consistent — `latest_message` is the genuinely new exposure.) Fix: drop `latest_message`, or path-sanitize it.
- **Minor — #714** `src/tools/live_config_preflight.py`: surfaces raw `str(path)` (expanduser-expanded) at `:233,239,248,254,261,301` in output the PR calls "operator evidence", leaking the deploy path/username. Fix: home-collapse the `config_path`/issue `path` values.

**Suggestion:** a single shared path-display helper applied uniformly to shareable tool output would close this whole class — it has now recurred in three consecutive tool PRs.

---

### Theme B — debug-profile enrichment: exception-isolation & scoping

(Directly relevant to the #723 family. For reference, #723 and #720 build their debug payload **inside** the wrapped/`_unchecked` emitter — the correct, exception-isolated pattern. The two issues below are where earlier slices diverged.)

- **Important — #718** `src/passivbot.py:~15361` (`rust_input_symbol_debug_sample`) and `:~15410` (`rust_output_order_debug_sample`): both samplers are built as **argument expressions in the caller frame**, gated by `if rust_debug_profile_enabled`, *outside* the emitter's `try/except`. The inner `try` at that call site wraps only `pbr.compute_ideal_orders_json`, and this runs on the order cancel/create path (`calc_ideal_orders_orchestrator` → reconciler). The emitter wrappers (`emit_rust_orchestrator_*`, `emit_event`) protect only the *emit*, not argument construction — so a builder exception (only when a debug profile is enabled) would propagate into order execution. The builders are defensively coded so a real crash is unlikely, but it breaks the codebase's "all event enrichment is best-effort" invariant. Fix: move the sampling inside the `_unchecked` emitter body, or wrap each builder in `try/except → None`.
- **Important — #720** `src/live/event_emitters.py:2177-2179`: EMA enrichment is gated on `live_event_debug_profile_enabled(bot,"ema") or live_event_debug_profile_enabled(bot,"candles")`, with a hardcoded `debug_profile="ema"` marker. Enabling **only** the `candles` profile silently attaches EMA-readiness debug to `ema.unavailable` events (component `ema.bundle`, unrelated to candles). Looks like a copy-paste from another slice; the `candles` branch is untested (the new test uses `("ema",)`). Fix: drop the `or … "candles"` clause.

---

### Standalone correctness

- **Important — #722** `src/tools/cache_integrity_doctor.py`: the candle-coverage evidence never compares mask length (`row_count`) against the canonical expected length `rows_in_month(year,month,tf)` from `ohlcv_store` (the file doesn't import it; it re-derives `_timeframe_interval_ms`/`_month_start_ms` at `:17`/`:32`), and gaps are interior-only (`np.diff(valid_indices)` at `:277`). A truncated / short `.valid.npy` — a real corruption mode (partial write, disk-full, interrupted download) — is loaded as-is and reports `gap_count=0` with `evidence="coverage_observed"` (`:407`), i.e. **false trust in a half-empty cache file**, with leading/trailing missing rows also uncounted. Fix: use `rows_in_month()` as the expected-length authority — emit a `coverage_length_mismatch` issue / distinct evidence label and treat trailing shortfall as a boundary gap. (Reusing the canonical `ohlcv_store` helpers also removes the re-implementation drift risk.)
- **Minor — #714** `_hsl_side_report` / `_forager_side_report` read only the *grouped* bot form (`side_config["hsl"]`, `side_config["risk"]["n_positions"]`). A valid **flat** bot config yields `n_positions=None`, `forager_present=false`, and null HSL fields silently — wrong risk facts, no warning. Low probability (shipped examples are grouped). Fix: resolve via `get_grouped_bot_value` (handles both forms).

---

### Lower priority / defense-in-depth (noting only)

- **#717** `src/tools/hsl_startup_preview.py:180`: the `HSL_DATA_KEYS` allowlist passes allowlisted values verbatim without scalar-type coercion; a non-scalar planted in an allowlisted key would surface unredacted. Not a live path (emitters always write float scalars) — defense-in-depth only. Otherwise this tool audited clean (no HSL-math re-implementation; read-only; secrets allowlisted).
- **#718** observations: `LIVE_EVENT_DEBUG_PROFILES` accepts profiles that were silent no-ops at merge time (only `rust` consumed then); and an invalid `PASSIVBOT_LIVE_EVENT_DEBUG_PROFILES` value raises a raw `ValueError` out of `Passivbot.__init__` (hard startup crash) rather than a clean rejection.

Tests for every audited PR pass at their merged SHAs; none of the above is a test regression. Net trading-safety risk is low (the one path-into-trading concern, #718-A, is gated behind an opt-in debug profile and defensively coded), but the path-hygiene leaks (Theme A) and the #722 false-coverage gap are worth a follow-up pass.

PR #818, opened 2026-06-28
Read explicit OHLCV source dirs directly

msei99:
Summary
- Read explicit backtest.ohlcv_source_dir trees directly for backtest/optimizer HLCV prep instead of importing raw candles into caches/ohlcvs first.
- Apply the same direct-source behavior in combined mode, including BTC benchmark loading and candidate source reporting.
- Update docs/changelog and add regression coverage.

Tests
- /home/mani/software/venv_pb7/bin/python -m py_compile src/hlcv_preparation.py src/backtest.py
- /home/mani/software/venv_pb7/bin/python -m pytest tests/test_ohlcv_v2_prepare.py tests/test_hlcv_preparation.py tests/test_market_settings_sources_separation.py
- git diff --check


enarjord (commented):
**PR:** #818 — Read explicit OHLCV source dirs directly
**Head SHA:** `8cae65746e794ad4faa59aff0ba52c7b040efa7e`
**Base:** v8 (merge-base `9dee208e`)
**Scope:** backtest/optimizer OHLCV data-prep — `src/hlcv_preparation.py` +34, `src/backtest.py` (−10 net), tests +127, docs/CHANGELOG. Offline data-prep only (no live-trading path, no event pipeline).
**Verdict:** ✅ **Approve** — no blockers (2 non-blocking observations)

### Intent & architectural fit
This generalizes a previously narrow special-case into a root-cause fix, which is the right direction. Before, the direct source-dir preparer in `prepare_hlcvs_mss` was gated on `stock_perp_coins (xyz:*) AND has_explicit_source_dir` — a band-aid for the Hyperliquid stock-perp startup issue. #818 broadens it to `if has_explicit_source_dir:` for any coin set, with the clear rationale "explicit source dirs are caller-managed OHLCV stores; read them directly instead of importing them into PB7's v2 raw OHLCV store." That removes a coin-type-specific carve-out in favour of one consistent rule, and applies the same treatment to combined mode (per-coin candidates + BTC benchmark + candidate source reporting).

### Behavior preservation — verified
The new `use_v2_local` parameter is threaded through the whole combined chain (`prepare_hlcvs_combined` → `_prepare_hlcvs_combined_impl` → `_resolve_combined_coin` → `_load_combined_coin_candidates` → per-exchange loader, plus `_load_combined_btc_prices`) with **default `use_v2_local=True` / `ohlcv_source_dir=None`** at every hop. So any config WITHOUT an explicit `ohlcv_source_dir` keeps the exact prior behavior (v2-local path, hardcoded-`True` equivalent). The only semantic change is for configs that explicitly set `ohlcv_source_dir`. The `_has_explicit_ohlcv_source_dir` helper is defensive (isinstance/`or ""`/strip). Source-layer log labels (`v2_store`/`source_dir`/`remote`) now reflect the path actually taken.

### Tests
Ran the touched suites in a clean /tmp worktree: **`tests/test_hlcv_preparation.py` + `tests/test_ohlcv_v2_prepare.py` → 12 passed**, no rust dependency. New coverage hits both key paths: `test_load_combined_coin_candidates_can_bypass_v2_for_source_dir` (combined v2-bypass) and `test_prepare_hlcvs_mss_crypto_source_dir_uses_direct_prepare` (the broadening to crypto coins — which also documents that the `xyz:`-removal is intentional, not an oversight).

### Observations (non-blocking)
1. **Intentional behavior broadening — worth a changelog callout for existing users.** Any pre-existing config that set `backtest.ohlcv_source_dir` for a *non*-stock-perp run and relied on the old import-into-v2 behavior will now read the dir directly instead. The PR makes this intentional (title/body + the new crypto test), but a one-line note in the changelog/docs that "explicit source dirs are now always read directly" would help anyone who depended on the import side effect.
2. **BTC direct-load column assumption.** In `_load_combined_btc_prices` (the `not use_v2_local` branch), `btc_df.loc[:, ["timestamp", "close"]]` assumes both columns exist on the source-dir BTC frame. The `if not btc_df.empty` guard covers the empty case, but a malformed source-dir BTC frame missing those columns would `KeyError` rather than fall through to the next candidate exchange. Low-risk (standard OHLCV frames have both), but a `set(...).issubset(btc_df.columns)` guard mirroring the empty-check would make it fully fail-soft.

Net: a clean, well-tested generalization that is behavior-preserving for the default path and opt-in for the changed path. Approving.

vigilpbclaw-hash:
PR: #818
Head SHA: 8cae65746e794ad4faa59aff0ba52c7b040efa7e
Base SHA/branch: 9dee208e5dd7cce8013ae270254e871775d0f78c/master
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: This PR changes explicit `backtest.ohlcv_source_dir` from an import-into-v2-local-store path into a caller-managed direct-read source for both single-exchange and combined backtest/optimizer preparation. That is a good fit for the stated intent: it removes the prior stock-perp-specific exception and makes the rule depend on the explicit data-source contract instead of coin naming. Default configs without `ohlcv_source_dir` continue through strict v2 materialization.

Evidence reviewed:
- `src/backtest.py`: non-combined `prepare_hlcvs_mss()` now routes any explicit source dir through `prepare_hlcvs(..., skip_v2_local=True)` while preserving strict v2 for normal runs.
- `src/hlcv_preparation.py`: combined mode threads `use_v2_local=False` into candidate loading and BTC benchmark loading only when an explicit source dir is configured; source-layer reporting now distinguishes `source_dir` from `v2_store`/`remote`.
- Tests cover the new crypto source-dir dispatch and combined candidate bypass behavior, and existing source-dir loader tests cover real `.npy`/`.npz` loading/fallback behavior.
- Docs and CHANGELOG describe the user-facing behavior change.

Checks run:
- `python3 -m py_compile src/hlcv_preparation.py src/backtest.py` (passed)
- `/Users/nodus/passivbot/venv/bin/python -m pytest tests/test_ohlcv_v2_prepare.py tests/test_hlcv_preparation.py -q` (82 passed, 2 skipped; warnings only)
- `git diff --check origin/master...HEAD` (passed)

No trading/order-risk behavior changed; this is offline HLCV preparation/data-source plumbing. OK to merge from my review.

PR #839, opened 2026-06-29
Guard HSL equity replay with balance override

enarjord:
## Summary
- fail HSL unified/pside account-level equity replay before startup history reconstruction when `balance_override` is active
- leave coin-mode HSL untouched because it does not reconstruct account-level equity from `balance - realized_pnl_total`
- add VPS3 incident report documenting the false XMR panic path

## Why
VPS3 `ebybitsub03` ran Bybit XMR with `hsl_signal_mode=unified` and `-bo 1000`. HSL reconstructed `peak_strategy_equity ~= 1213` from overridden balance plus historical realized PnL, saw ~17.5% raw drawdown, and eventually emitted a panic close even though the account allocation was not actually near red drawdown.

## Tests
- `PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_realized_loss_gate.py -q`
- `PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_run_fake_live.py -k hsl_replay_scenarios_run_end_to_end -q`
- `PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/passivbot_hsl.py src/passivbot.py tests/test_realized_loss_gate.py`
- `git diff --check`

PR #840, opened 2026-06-29
Document HSL false-panic recovery follow-up

enarjord:
Docs-only follow-up to PR #839. Records the confirmed VPS3 false HSL panic state, clarifies that current v8 prevents recurrence by fail-loud guarding balance_override + account-level HSL replay, and adds a backlog item for explicit invalid-panic recovery/preflight design. No code or trading behavior changes.

PR #841, opened 2026-06-29
Preflight HSL balance override replay risk

enarjord:
Adds an offline live-config-preflight check for the unsafe contract fixed at runtime in PR #839: active balance_override plus HSL signal_mode unified/pside. Also adds --balance-override so operators can preflight an intended live command using -bo, not just a JSON config field. Coin-mode HSL remains allowed. Validation: PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_config_preflight.py -q; compileall src/tools/live_config_preflight.py tests/test_live_config_preflight.py; git diff --check; manual CLI smoke returned RC 1 with hsl_balance_override_account_level_replay_unsafe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- `passivbot tool live-config-preflight` now flags active
+  `balance_override` plus account-level HSL signal modes (`unified`/`pside`)
+  before live startup, including an optional `--balance-override` argument for
+  preflighting runs that will pass `-bo`.
 - Added a live HSL safety guard: `hsl_signal_mode=unified` or `pside` now
   fails before account-level equity replay when `balance_override` is active,
   preventing synthetic historical peaks from triggering false RED panic orders
--- a/src/tools/live_config_preflight.py
+++ b/src/tools/live_config_preflight.py
@@ -419,7 +419,7 @@ def _balance_override_report(
         }
     if isinstance(raw_value, bool):
         return {
-            "active": False,
+            "active": True,
             "source": source,
             "present": True,
             "value_type": type(raw_value).__name__,
@@ -429,7 +429,7 @@ def _balance_override_report(
         number = float(raw_value)
     except (TypeError, ValueError):
         return {
-            "active": False,
+            "active": True,
             "source": source,
             "present": True,
             "value_type": type(raw_value).__name__,
@@ -437,9 +437,10 @@ def _balance_override_report(
         }
     if not math.isfinite(number) or number <= 0.0:
         return {
-            "active": False,
+            "active": True,
             "source": source,
             "present": True,
+            "value": number if math.isfinite(number) else None,
             "value_type": type(raw_value).__name__,
             "status": "invalid",
         }
@@ -985,7 +986,7 @@ def build_live_config_preflight_report(
     if balance_override_report.get("status") in {"invalid", "invalid_bool"}:
         issues.append(
             _issue(
-                "warning",
+                "error",
                 f"balance_override_{balance_override_report['status']}",
                 "balance override is present but is not a positive finite number",
                 path=(
--- a/src/tools/live_config_preflight.py
+++ b/src/tools/live_config_preflight.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 from collections import Counter
 import json
+import math
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


vigilpbclaw-hash:
PR: #841
Head SHA: e2cfa13aeb3c50136c96cbc45748b16520be82fb
Base SHA/branch: a76a835f6d882bd70dfeb07a66288a41ccb7f11b/v8
Scope reviewed: full PR
Verdict: needs fix

- Important: `src/tools/live_config_preflight.py:438` / `src/tools/live_config_preflight.py:985` treats non-positive and boolean `balance_override` values as inactive warnings, so the preflight command can still return `ok=true`/exit 0 for `balance_override=0`, `-1`, `false`, or `'0'` with account-level HSL enabled. That does not match the runtime guard added in #839: runtime considers any non-`None` `self.balance_override` active (`src/passivbot_hsl.py:473`) and blocks unified/pside HSL replay before history initialization. I confirmed with a local smoke that those values currently produce only `balance_override_invalid*` warnings and `ok=True`. For this preflight to be a reliable offline check of the live-startup contract, invalid/non-positive override values should also make the report non-ok (or at least trigger the same HSL replay-blocking error whenever runtime would see an override), with tests for these cases.

- Minor: no separate finding.

Intent/architecture pass: the PR is trying to move the #839 fail-loud HSL/balance-override contract into the offline `live-config-preflight` tool, including launch-time `--balance-override`. That is the right abstraction: a read-only preflight should mirror the live startup safety gate and should not alter trading behavior. The implementation is close, but the override parser now uses a different definition of “active” than runtime, which creates a future operability trap where preflight can green-light a launch shape that live will reject.

Validation run:
- `git diff --check origin/v8...HEAD`
- `PYTHONPATH=src python3 -m pytest tests/test_live_config_preflight.py -q` (17 passed)
- `python3 -m compileall -q src/tools/live_config_preflight.py tests/test_live_config_preflight.py`
- merge dry-run in a detached `origin/v8` worktree: `git merge --no-commit --no-ff origin/pr-841` (automatic merge succeeded)
- targeted smoke for `balance_override` values `0`, `-1`, `False`, and `'0'` with HSL enabled (all produced `ok=True` with warning-only issues)

enarjord (commented):
**PR:** #841 — Preflight HSL balance override replay risk
**Head SHA:** `e2cfa13aeb3c50136c96cbc45748b16520be82fb`
**Base:** v8 (merge-base `a76a835f` = current tip)
**Scope:** offline tooling — `src/tools/live_config_preflight.py` +128, tests +86, CHANGELOG +4. No runtime/trading code.
**Verdict:** ✅ **Approve** — no findings (2 tiny non-blocking notes)

Implements the offline-preflight half of the #840 backlog item `0a`: it lets operators catch the unsafe (balance_override + account-level HSL) contract *before* starting live, complementing #839's runtime fail-loud guard. Verified:

- **Offline / no runtime change.** Entirely within `live_config_preflight.py` (a config-validation tool) + tests; no `passivbot.py`/`passivbot_hsl.py` trading-path change.
- **Mirrors the #839 runtime condition.** The check fires exactly when HSL is enabled AND `balance_override` is active AND `live.hsl_signal_mode ∈ {unified, pside}` — emitting `hsl_balance_override_account_level_replay_unsafe` with an accurate message ("runtime will fail before replay…"). Coin-mode is correctly left allowed. Wired into both `build_live_config_preflight_report` and the diff-report's `_cache_readiness_report`.
- **`--balance-override` arg.** `_balance_override_report` resolves the override from the launch argument or `live.balance_override`, so an operator can preflight an intended `-bo 1000` command, not just a JSON field. Invalid/non-positive/non-finite overrides are separately flagged as `balance_override_invalid` (warning).
- **Value-safe / path-hygiene.** It's an operator-local tool echoing the operator's own config/override value (not a shareable live surface); `_user_safe_display_path` is preserved for paths.
- **Tests:** ran `tests/test_live_config_preflight.py` in a clean /tmp worktree — **17 passed**, covering the config-field flag, the `--balance-override` launch flag, coin-mode being allowed, and the CLI returning a nonzero exit code on the unsafe contract.

**Tiny non-blocking notes:**
1. The unsafe-contract condition is re-implemented here against raw config dicts (the runtime guard `_equity_hard_stop_validate_balance_source_for_history_replay` operates on a bot instance, so direct reuse isn't trivial). Worth a comment cross-referencing #839 so the two conditions stay in sync if the runtime rule ever changes.
2. Minor edge divergence: a non-positive/non-finite `balance_override` is treated as `active: False` here (flagged as `balance_override_invalid`) whereas the #839 runtime treats any non-None value as active. Both warn the operator, so this is not a gap — just a slightly different label for an unrealistic value.

Clean, well-tested preflight that closes the "missed in a manual restart" gap from the incident report. Approving.

enarjord (commented):
**Reviewer:** Claude (automated continuous reviewer) — re-review (amended head)
**PR:** #841 — Preflight HSL balance override replay risk
**Head SHA:** `bf2bab345c235794e54dcae3fdf643114d7e5ac9` (was `e2cfa13a`)
**Base:** v8 (merge-base `53b8b40a` = current tip, post-#840)
**Verdict:** ✅ **Approve** — no findings

The amendment ("Align preflight balance override semantics") directly addresses Note #2 from my prior review and tightens the check. Verified:

- **Active-detection now mirrors the #839 runtime exactly.** `_balance_override_report` now returns `active: True` for *any* present, non-`None`/non-`""` `balance_override` — including bool / unparseable / non-finite / `≤0` (each also tagged `status: invalid`/`invalid_bool`). That matches the runtime, where `balance_override = None if raw in (None, "") else float(raw)` makes any present value non-`None` and thus trips the #839 guard. So the preflight now flags `hsl_balance_override_account_level_replay_unsafe` for the same input set the runtime fails on, eliminating the prior divergence where a non-positive override slipped past the unsafe-replay flag.
- **Invalid overrides escalated to `error`** (was `warning`), so the CLI exits nonzero on them too — appropriate, since an invalid override is both unusable and (with unified/pside HSL) unsafe.
- **Still offline tooling, no runtime/trading change** — entirely within `live_config_preflight.py` + tests; coin-mode HSL remains allowed; `--balance-override` continues to let operators preflight a `-bo` command; value-safe (operator-local, `_user_safe_display_path` for paths).
- **Tests:** ran `tests/test_live_config_preflight.py` in a clean /tmp worktree — **22 passed** (up from 17; new cases cover the bool/invalid/non-positive override now being flagged as active+error and the unsafe-replay flag firing accordingly).

My approve stands and is strengthened by the alignment. (Prior Note #1 — a comment cross-referencing the #839 runtime condition so the two stay in sync — remains a nice-to-have, not a blocker.)

vigilpbclaw-hash:
PR: #841
Head SHA: bf2bab345c235794e54dcae3fdf643114d7e5ac9
Base SHA/branch: 53b8b40abe850b009261dd4f07eae16391d1950a/v8
Scope reviewed: delta e2cfa13aeb3c50136c96cbc45748b16520be82fb..bf2bab345c235794e54dcae3fdf643114d7e5ac9
Verdict: approve

Previous Hermes review for head `e2cfa13aeb3c50136c96cbc45748b16520be82fb` is superseded by this head SHA. The delta review included the fix commit plus current full-PR context against `origin/v8` because the branch was rebased onto newer `v8` and the raw old..new range also contains base-branch changes.

- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: the PR is still an offline preflight mirror of the live HSL fail-loud contract for `balance_override` with account-level HSL (`unified`/`pside`). The new delta aligns the preflight definition of active override with runtime (`balance_override is not None`) by treating invalid/non-positive present values as active and erroring, while preserving the separate invalid-value diagnostic. That is the right design for this operator-safety tool: fail before live startup rather than green-lighting a shape that runtime would reject. It remains read-only/offline and does not alter live trading behavior.

Validation run:
- `git diff --check origin/v8...origin/pr-841`
- `PYTHONPATH=src /tmp/passivbot-pr841-venv/bin/python -m pytest tests/test_live_config_preflight.py -q` — 22 passed
- `PYTHONPATH=src /tmp/passivbot-pr841-venv/bin/python -m compileall -q src/tools/live_config_preflight.py tests/test_live_config_preflight.py`
- merge dry-run in detached `origin/v8` worktree: `git merge --no-commit --no-ff origin/pr-841` — automatic merge succeeded

PR #835, opened 2026-06-28
Emit realized loss gate block events

enarjord:
Adds a structured risk.realized_loss_gate_blocked event next to the existing throttled realized-loss gate warning. No trading decision changes; emission is best-effort and routed off console/text by default. The read-only live-performance-report risk_activity summary includes the new event type using envelope labels only.\n\nValidation:\n- pytest tests/test_realized_loss_gate.py -q -> 36 passed\n- pytest tests/test_live_performance_report.py -q -> 46 passed\n- pytest tests/test_live_event_bus.py -q -> 33 passed\n- py_compile touched Python files\n- git diff --check\n\nNote: initial pytest collection hit the local stale Rust-extension guard; rebuilt/restamped the extension, then tests passed in fresh Python processes.

PR #846, opened 2026-06-29
Emit create skip events for market-distance guard

enarjord:
## Summary
- emit off-console structured execution.create_skipped events when the pre-create market-distance guard skips far-from-market limit creates
- add a stable limit_order_create_market_distance reason code and registry docs entry
- extend the existing pre-create distance guard test to verify bounded event groups/samples alongside the text log

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_passivbot_balance_split.py::test_pre_create_snapshot_filter_skips_far_limit_order_creations tests/test_live_event_registry_docs.py tests/test_live_event_bus.py::test_live_event_reason_code_registry_values_are_unique_and_query_safe -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/live/market_data.py src/live/event_bus.py tests/test_passivbot_balance_split.py
- git diff --check

### DIFF
--- a/docs/ai/live_event_registry.md
+++ b/docs/ai/live_event_registry.md
@@ -152,6 +152,7 @@ across time.
 - `hsl_balance_override_account_level_replay_unsafe`
 - `initial_entry_distance_gate`
 - `length_mismatch`
+- `limit_order_create_market_distance`
 - `low_balance`
 - `min_effective_cost_blocked`
 - `new_fill`
--- a/src/live/event_bus.py
+++ b/src/live/event_bus.py
@@ -193,6 +193,7 @@ class ReasonCodes:
     FILL_CACHE_READY = "fill_cache_ready"
     INITIAL_ENTRY_DISTANCE_GATE = "initial_entry_distance_gate"
     LENGTH_MISMATCH = "length_mismatch"
+    LIMIT_ORDER_CREATE_MARKET_DISTANCE = "limit_order_create_market_distance"
     LOW_BALANCE = "low_balance"
     MIN_EFFECTIVE_COST_BLOCKED = "min_effective_cost_blocked"
     NEW_FILL = "new_fill"
--- a/src/live/market_data.py
+++ b/src/live/market_data.py
@@ -8,6 +8,7 @@ from collections import Counter
 from typing import Iterable
 
 from config.access import get_optional_live_value
+from live.event_bus import EventTypes, ReasonCodes
 from live.market_snapshot import MarketSnapshot
 from utils import utc_ms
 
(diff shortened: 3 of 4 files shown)


vigilpbclaw-hash:
PR: #846
Head SHA: 6d2717da6cda8e802f322466d922f2a99fdad52f
Base SHA/branch: e6fc95121d2ecb0295c9718b3b43a2af2f8da786/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this PR is observability-only for the pre-create market-distance guard, adding a stable `execution.create_skipped` reason code plus bounded group/sample payloads while keeping the existing order-filtering behavior and text-log throttling intact. This fits the existing live event pipeline abstractions (`_emit_execution_create_filter_event`, default non-console route for create-skipped events) and avoids adding new exchange-action gating or changing trading behavior. Event emission is best-effort through the existing emitter, which catches emission failures, so sink/pipeline failures should not alter live execution.

Evidence reviewed:
- Diff: `src/live/market_data.py`, `src/live/event_bus.py`, `docs/ai/live_event_registry.md`, `tests/test_passivbot_balance_split.py`.
- Merge check: PR head merges cleanly into current `origin/v8` (`e6fc95121d2ecb0295c9718b3b43a2af2f8da786`).
- Payload hygiene: emitted samples are bounded (`groups` 12, `sample` 8), use symbol/side/type/price/market distance only, and the route remains structured/off-console by default.
- Duplicate guard: no previous Hermes Continuous Reviewer review found for this head SHA before posting.

Validation run:
- `git diff --check origin/v8...origin/pr-846` — passed.
- `PYTHONPATH=src python -m pytest tests/test_passivbot_balance_split.py::test_pre_create_snapshot_filter_skips_far_limit_order_creations tests/test_live_event_registry_docs.py tests/test_live_event_bus.py::test_live_event_reason_code_registry_values_are_unique_and_query_safe -q` — passed (`6 passed`; one pre-existing `candlestick_manager.py` SyntaxWarning about `return` in `finally`).
- `PYTHONPATH=src python -m compileall src/live/market_data.py src/live/event_bus.py tests/test_passivbot_balance_split.py` — passed.

enarjord (commented):
**PR:** #846 — Emit create skip events for market-distance guard
**Head SHA:** `6d2717da6cda8e802f322466d922f2a99fdad52f`
**Base:** v8 (merge-base `e6fc9512` = current tip)
**Scope:** `src/live/market_data.py` (+44, the emit), `src/live/event_bus.py` (+1 reason code), registry doc (+1), `tests/test_passivbot_balance_split.py` (+50). Reuses the existing `execution.create_skipped` event type with a new `limit_order_create_market_distance` reason code.
**Verdict:** ✅ **Approve** — no blockers; 2 minor non-blocking notes.

Continues the gate-emitter family (cf. #835/#837/#838/#842): surfaces the pre-create market-distance skip — previously visible only as a throttled text log — as a structured off-console event. Verified against the LIVE-PATH EMITTER contract:

- **Observability-only (no trading change) — verified.** The skip decision happens entirely in the filter loop (`if dist > threshold: skipped.append(...); continue`) which partitions orders into `kept`/`skipped` *before* `_log_limit_order_distance_skips` is called. The new emit runs on the already-decided `skipped` list, returns nothing, and does not touch `kept` — so which orders are created is unchanged.
- **Exception-isolated — verified.** `emit_execution_create_filter_event` wraps its entire body (id resolution, payload build, debug-profile, `_emit_live_event`) in `try/except Exception → logging.debug`, and the call site is reached via `getattr(bot, "_emit_execution_create_filter_event", None)` + `callable(...)`, so a missing method or a telemetry failure cannot disturb the create path.
- **Off-console — verified.** `EXECUTION_CREATE_SKIPPED` is routed `EventRoute(console=False, text=False)` (pre-existing), and the payload carries `suppressed_text_log = not should_info` so the event mirrors the existing text-throttle state.
- **Value-safe for shareable reports — verified.** The `data` carries raw `price`/`market_price`/`distance` (log-parity values), which is acceptable in the off-console NDJSON. Confirmed nothing shareable reads this event's `data`: neither `performance_report.py`, `smoke_report.py`, `event_query.py`, nor `src/tools/` consume `execution.create_skipped`; the report-facing surfaces read envelope labels only (`status`/`reason_code`/`symbol`/`component`). The console summary (`_console_create_filter_summary`) is unreachable for this type since the route is `console=False`.
- **Bounded — verified.** Group sample capped at `most_common(12)`, order sample at `skipped[:8]`; the emitter additionally caps `symbols[:12]` with `symbols_count` + `symbols_truncated`. The true magnitude is never hidden because `order_count = len(skipped)` carries the full count.
- **Registry-documented — verified.** New reason code added to both `ReasonCodes` and `docs/ai/live_event_registry.md`; event type pre-existing/documented. The registry-uniqueness/docs-sync tests pass.

**Tests:** ran in a clean /tmp worktree at the PR head — `test_passivbot_balance_split.py` + `test_live_event_registry_docs.py` + `test_live_event_bus.py` → **227 passed, 0 failed**. The extended `test_pre_create_snapshot_filter_skips_far_limit_order_creations` asserts the bounded event groups/sample alongside the text log.

**Minor (hardening, non-blocking).** The sample-building in `_log_limit_order_distance_skips` (`float(order.get("price"))`, `float(item["market_price"])`, `float(item["dist"])` and the two comprehensions) runs *before* the call into the exception-isolated emitter, so it isn't itself wrapped. In practice it can't raise — every entry in `skipped` already passed `price = float(order["price"])` / `market_price = float(snapshot.last)` / `dist = float(...)` in the filter loop, so these are re-conversions of known-good floats. Still, if you want full belt-and-suspenders on the create path, moving the sample assembly inside the emitter's `try` (or guarding it) would close the theoretical gap.

**Minor (observation, non-blocking).** The 12-group / 8-order sample caps have no explicit `groups_truncated` / `sample_truncated` flag (unlike `symbols_truncated`). Low impact since `order_count` conveys the true total, but a truncation flag would make "the sample is partial" self-describing for a consumer.

Net: a clean, well-scoped, contract-compliant observability addition that makes a previously log-only entry gate queryable without altering execution. Approving.

PR #858, opened 2026-06-29
Instrument HSL history build progress

enarjord:
## Summary
- add opt-in structured HSL replay progress events from balance/equity history assembly
- distinguish history inputs loaded, price-history fetch start/completion, and timeline replay start/completion before the existing HSL history-loaded event
- opt HSL account-level and coin-mode replay callers into the progress events
- register new reason codes and catch up the live logging progress ledger through PRs #856/#857 + VPS5 smoke

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_passivbot_balance_split.py::test_balance_equity_history_paces_replay_candle_fetches tests/test_passivbot_balance_split.py::test_balance_equity_history_skips_unsupported_closed_historical_symbols tests/test_live_event_registry_docs.py tests/test_hsl_coin_mode.py -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall -q src/passivbot.py src/passivbot_hsl.py src/live/event_bus.py tests/test_passivbot_balance_split.py tests/test_hsl_coin_mode.py tests/test_unstucking_safeguards.py tests/test_live_event_registry_docs.py
- git diff --check

## Notes
- observability-only; the new get_balance_equity_history keyword defaults off and only HSL replay callers opt in
- one broader legacy account-level HSL safeguard test still fails on the existing unified-mode fixture missing unrealized_pnl_total; not changed by this slice

PR #861, opened 2026-06-29
Fix live event query bot path filtering

enarjord:
## Summary
- Match live-event-query --bot-id filters against the monitor path-derived bot id when an event lacks nested ids.bot_id.
- Keep emitted event IDs unchanged; the path-derived id is only a query fallback.
- Add a regression for HSL events under monitor/gateio/gateio_01/events/current.ndjson without event-level bot_id.

## Tests
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/live/event_query.py tests/test_live_event_query.py
- git diff --check
- rg -n "except Exception|return_exceptions=True|\.get\([^\n]*,\s*(0|0\.0|None|False|\{\}|\[\])\)" src/live/event_query.py tests/test_live_event_query.py

### DIFF
--- a/src/live/event_query.py
+++ b/src/live/event_query.py
@@ -111,6 +111,20 @@ def _event_ids(live_event: dict[str, Any]) -> dict[str, Any]:
     return out
 
 
+def _monitor_path_bot_id(path: Path) -> str | None:
+    """Return monitor path-derived bot id for .../<exchange>/<user>/events/*.ndjson."""
+    if path.parent.name != "events":
+        return None
+    try:
+        user = path.parent.parent.name
+        exchange = path.parent.parent.parent.name
+    except IndexError:
+        return None
+    if not exchange or not user:
+        return None
+    return f"{exchange}/{user}"
+
+
 def _normalize_filter_values(values: str | Iterable[str] | None) -> set[str]:
     if values is None:
         return set()
@@ -1047,7 +1061,9 @@ def build_event_report(
                         record_event_type, event_type_filter
                     )
                     cycle_matches = cycle_id is None or record_cycle_id == str(cycle_id)
-                    bot_matches = _filter_matches(ids.get("bot_id"), bot_filter)
+                    bot_matches = _filter_matches(
+                        ids.get("bot_id") or _monitor_path_bot_id(path), bot_filter
+                    )
                     snapshot_matches = _filter_matches(
                         ids.get("snapshot_id"), snapshot_filter
                     )
--- a/tests/test_live_event_query.py
+++ b/tests/test_live_event_query.py
@@ -543,6 +543,58 @@ def test_event_query_filters_by_remaining_event_ids(tmp_path):
     }
 
 
+def test_event_query_bot_filter_falls_back_to_monitor_path(tmp_path):
+    events_dir = tmp_path / "monitor" / "gateio" / "gateio_01" / "events"
+    _write_ndjson(
+        events_dir / "current.ndjson",
+        [
+            _monitor_row(
+                event_type="hsl.status",
+                cycle_id="cy_1",
+                seq=1,
+                ts=1000,
+                symbol="ZEC/USDT:USDT",
+                pside="long",
+            ),
+            _monitor_row(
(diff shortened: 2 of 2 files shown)


vigilpbclaw-hash:
PR: #861
Head SHA: c0acca7baefe6a42d9da4e0645041b5946151a5f
Base SHA/branch: 221a75d324383cce25573e27e8fdc8d77151996d/v8
Scope reviewed: full PR
Verdict: approve

Findings: none.

Intent/architecture pass: this PR is narrowly scoped to make `live_event_query --bot-id` match monitor-routed events that do not carry nested `ids.bot_id`, without changing emitted event IDs. Using the canonical monitor path shape (`.../<exchange>/<user>/events/*.ndjson`) as a query-only fallback fits the existing discovery contract and keeps event emission behavior untouched. Existing nested `ids.bot_id` remains authoritative when present, so historical/id-rich records retain their current behavior.

Validation run:
- `git diff --check origin/v8...HEAD`
- `PYTHONPATH=src python3 -m pytest tests/test_live_event_query.py -q` — 26 passed (one existing pytest config warning about `asyncio_mode` under this interpreter)
- `PYTHONPATH=src python3 -m compileall -q src/live/event_query.py tests/test_live_event_query.py`
- Diff scan for debug statements, secrets-like patterns, and conflict markers — no matches

Minor notes: none blocking.

enarjord (commented):
**PR:** #861 — Fix live event query bot path filtering
**Head SHA:** `c0acca7baefe6a42d9da4e0645041b5946151a5f`
**Base:** v8 (merge-base `221a75d3` = current tip)
**Scope:** `src/live/event_query.py` (+18, `_monitor_path_bot_id` + one filter line), `tests/test_live_event_query.py` (+52 regression). Offline read-only query tooling.
**Verdict:** ✅ **Approve** — no blockers, no findings.

Fixes a real `live-event-query --bot-id` gap: events that don't carry a nested `ids.bot_id` (e.g. the HSL replay events, which `_emit_hsl_replay_event` may emit without one — exactly the new #858/#860 progress events) were unmatchable by `--bot-id`. The fix derives the bot id from the monitor path as a query-only fallback.

- **Correct, fallback-only.** `bot_matches = _filter_matches(ids.get("bot_id") or _monitor_path_bot_id(path), bot_filter)` — a present nested `bot_id` is used unchanged; the path-derived id only fills in when the nested one is absent/falsy. No event-level id is overridden.
- **Format consistent with the existing convention.** `_monitor_path_bot_id` returns `f"{exchange}/{user}"` from `.../<exchange>/<user>/events/*.ndjson`, which is byte-for-byte the same bot-id format the smoke report already uses (`smoke_report.py:223` → `f"{exchange}/{user}"`), so `--bot-id` values that match the smoke report's bot ids now also match these events.
- **Robust path parsing.** Guards on `path.parent.name == "events"` and returns `None` on empty exchange/user (shallow paths yield empty `.name` components, caught by `if not exchange or not user`); the `try/except IndexError` is harmless belt-and-suspenders (pathlib `.parent` doesn't raise). No file I/O — only pathlib name components.
- **No emit-path change / read-only.** Only `build_event_report`'s filter is touched; emitted event IDs are unchanged (as the PR states), and the query tool remains offline/read-only. No new data is surfaced — the path-derived id is exchange/user labels, not financial — so no value-safety impact.

**Tests:** ran in a clean /tmp worktree at the PR head — `test_live_event_query.py` → **26 passed, 0 failed**. The new regression covers HSL events under `monitor/gateio/gateio_01/events/current.ndjson` with no event-level `bot_id` being matched by the `--bot-id` filter.

**Architectural fit.** Sensible call to fix this query-side (path fallback) rather than backfilling a nested `bot_id` at every emit site — it's lower-risk (no trading/emit-path change) and immediately makes the new HSL progress events queryable by bot, which is what operators need to diagnose the per-bot price-history fetch latency. Approving.

PR #872, opened 2026-06-29
Summarize HSL flat anchor sources

enarjord:
## Summary
- add a bounded risk_events.hsl_flat_finalization_anchors aggregate to live-smoke-report
- count hsl.red_finalized_without_order stop-event anchor sources and current-time fallback usage
- project the aggregate through full, --summary, and --brief smoke outputs

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_smoke_report.py -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_smoke_report.py::test_live_smoke_report_summarizes_recent_risk_events tests/test_live_smoke_report.py::test_live_smoke_report_cli_can_emit_concise_summary tests/test_live_smoke_report.py::test_live_smoke_report_cli_can_emit_brief_summary -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall -q src/live/smoke_report.py tests/test_live_smoke_report.py
- git diff --check

Observability/report-only: no event producers, exchange calls, cache mutation, readiness gates, console routing, order/risk logic, or trading behavior changed.

PR #871, opened 2026-06-29
Record HSL anchor provenance progress

enarjord:
## Summary
- record PR #870 review, deploy, and VPS5 smoke evidence in the live logging progress ledger
- update the current logging-overhaul head to v8@b6cd3b9e

## Validation
- git diff --check

Docs-only; no runtime code or trading behavior changed.

PR #870, opened 2026-06-29
Record HSL flat finalization anchor source

enarjord:
## Summary
- add stop-event anchor provenance to hsl.red_finalized_without_order events
- distinguish panic_fill, current_time_fallback, and provided_stop_event anchors without changing stop/cooldown behavior
- project the new bounded provenance fields in live-smoke-report risk_events

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_smoke_report.py
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_hsl_coin_mode.py tests/test_unstucking_safeguards.py::test_hard_stop_finalize_red_stop_uses_latest_panic_fill_timestamp
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/passivbot_hsl.py src/passivbot.py src/live/smoke_report.py tests/test_hsl_coin_mode.py tests/test_live_smoke_report.py
- git diff --check

Observability-only: no order creation, cooldown math, latch writes, exchange calls, readiness gates, or console routing changed.

PR #777, opened 2026-06-27
Document live performance readiness goals

enarjord:
## Summary
- add a performance/readiness checklist for live bot timing, HSL protective readiness, HSL replay speed, checkpointing, startup, and resource profiling
- ground the checklist in current VPS5 smoke/monitor evidence, including coin-HSL replay latency and runtime cycle timing samples
- keep this as docs-only planning; no trading behavior or runtime code changes

## Validation
- git diff --check -- docs/plans/live_performance_readiness_goals.md

PR #873, opened 2026-06-29
Filter live event queries by data fields

enarjord:
## Summary
- add --data-eq key=value to passivbot tool live-event-query
- support repeated top-level event data equality filters in build_event_report
- include data_eq in query/cycle filter metadata and compose it with existing event/tag/symbol/id filters

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py::test_event_query_filters_by_top_level_data_fields tests/test_live_event_query.py::test_event_query_rejects_malformed_data_eq_filter tests/test_live_event_query.py::test_live_event_query_cli_accepts_data_eq_filter tests/test_live_event_query.py::test_live_event_query_cli_rejects_malformed_data_eq_filter -q
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall -q src/live/event_query.py src/tools/live_event_query.py tests/test_live_event_query.py
- git diff --check
- silent-handling scan on touched files

Query/tooling-only: no event producers, exchange calls, cache mutation, readiness gates, console routing, order/risk logic, or trading behavior changed.

### DIFF
--- a/src/live/event_query.py
+++ b/src/live/event_query.py
@@ -145,6 +145,54 @@ def _filter_matches(value: Any, filter_values: set[str]) -> bool:
     return not filter_values or (value is not None and str(value) in filter_values)
 
 
+def _normalize_data_eq_filters(
+    values: str | Iterable[str] | None,
+) -> dict[str, set[str]]:
+    if values is None:
+        return {}
+    if isinstance(values, str):
+        raw_values = [values]
+    else:
+        raw_values = list(values)
+    out: dict[str, set[str]] = defaultdict(set)
+    for raw_value in raw_values:
+        text = str(raw_value)
+        if "=" not in text:
+            raise ValueError("data_eq filters must use key=value")
+        key, expected = text.split("=", 1)
+        key = key.strip()
+        if not key:
+            raise ValueError("data_eq filters must include a non-empty key")
+        out[key].add(expected.strip())
+    return {key: set(values) for key, values in sorted(out.items())}
+
+
+def _data_value_matches(value: Any, expected_values: set[str]) -> bool:
+    candidates = {str(value)}
+    if isinstance(value, bool):
+        candidates.add(str(value).lower())
+    elif value is None:
+        candidates.add("null")
+    return bool(candidates.intersection(expected_values))
+
+
+def _data_filters_match(
+    live_event: dict[str, Any],
+    data_eq_filters: dict[str, set[str]],
+) -> bool:
+    if not data_eq_filters:
+        return True
+    data = live_event.get("data")
+    if not isinstance(data, dict):
+        return False
+    for key, expected_values in data_eq_filters.items():
+        if key not in data:
+            return False
+        if not _data_value_matches(data.get(key), expected_values):
+            return False
+    return True
+
+
 def _filter_report(
     *,
     cycle_id: str | None,
@@ -161,6 +209,7 @@ def _filter_report(
     reason_codes: set[str],
     statuses: set[str],
     tags: set[str],
+    data_eq_filters: dict[str, set[str]],
     since_ms: int | None = None,
     until_ms: int | None = None,
 ) -> dict[str, Any]:
@@ -197,6 +246,10 @@ def _filter_report(
         filters["statuses"] = sorted(statuses)
     if tags:
         filters["tags"] = sorted(tags)
+    if data_eq_filters:
+        filters["data_eq"] = {
+            key: sorted(values) for key, values in sorted(data_eq_filters.items())
+        }
     return filters
 
 
@@ -856,6 +909,7 @@ def build_event_report(
     reason_code: str | Iterable[str] | None = None,
     status: str | Iterable[str] | None = None,
     tag: str | Iterable[str] | None = None,
+    data_eq: str | Iterable[str] | None = None,
     since_ms: int | None = None,
     until_ms: int | None = None,
     limit: int = 200,
@@ -934,6 +988,7 @@ def build_event_report(
     reason_code_filter = _normalize_filter_values(reason_code)
     status_filter = _normalize_filter_values(status)
     tag_filter = _normalize_filter_values(tag)
+    data_eq_filters = _normalize_data_eq_filters(data_eq)
     has_non_cycle_filter = any(
         (
             event_type_filter,
@@ -949,6 +1004,7 @@ def build_event_report(
             reason_code_filter,
             status_filter,
             tag_filter,
+            data_eq_filters,
         )
     )
     trace_without_cycle = (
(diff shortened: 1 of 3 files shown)


vigilpbclaw-hash:
PR: #873
Head SHA: dc30a4c39bef137babe9400ef13ea43d49440188
Base SHA/branch: c45537c1e15979358a581bd98cbc52ddc3bc46d8/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none
- Important: none
- Minor: none

Intent/architecture pass: this PR is narrowly scoped to add an observability/query filter for top-level live-event `data` fields. The implementation keeps the behavior in `src/live/event_query.py` and only wires the CLI through `src/tools/live_event_query.py`, so it stays query/tooling-only and does not alter event production, exchange calls, order/risk logic, readiness gates, or live trading behavior. The filter composes with the existing event/id/tag/time filters and reports its normalized criteria in query/cycle metadata, which fits the current `build_event_report` abstraction.

Validation run by reviewer:
- `git diff --check origin/v8...HEAD` — passed
- Merge dry-run into current `origin/v8` via `git merge --no-commit --no-ff origin/pr-873` — clean
- `PYTHONPATH=src python3 -m pytest tests/test_live_event_query.py -q` — passed: 30 passed, 1 pytest config warning about unknown `asyncio_mode`
- `PYTHONPATH=src python3 -m compileall -q src/live/event_query.py src/tools/live_event_query.py tests/test_live_event_query.py` — passed

enarjord (commented):
**PR:** #873 — Filter live event queries by data fields
**Head SHA:** `fc63a3ae3f79ed304446a254c235ac45b5ee10ce`
**Base:** v8 (merge-base `41db9b93` = #871 tip; branched before #872 merged, but #872 only touched `smoke_report.py` so no overlap — own-delta is clean)
**Scope:** local event-query tooling — `src/live/event_query.py` (+60), `src/tools/live_event_query.py` (+64/−26), `tests/test_live_event_query.py` (+122). Read-only NDJSON inspection CLI; no live/trading path, no event producers.
**Verdict:** ✅ **Approve** — no blockers (1 scope note)

This adds a `--data-eq key=value` filter to the `live-event-query` CLI so operators can narrow compact records by top-level event `data` fields — e.g. `--data-eq stop_event_anchor_source=current_time_fallback` to find exactly the #870 flat-finalizations that fell back to current time. A genuinely useful diagnostic for the #777/#802 work. Reviewed against the read-only / value-safe / bounded contract.

### Read-only & correct
Pure additive filtering over already-read NDJSON; no mutation, exchange calls, producers, or trading logic. The filter semantics are sound:
- `_normalize_data_eq_filters` parses `key=value` into `dict[str, set[str]]` (repeated keys OR their values), rejecting input with no `=` or an empty key via `ValueError`.
- `_data_filters_match` is **AND across keys, OR within a key's values**, and fails closed (returns `False`) when `data` isn't a dict or a required key is absent.
- `_data_value_matches` normalizes sensibly: bool → `{"True","true"}`, `None` → `{"None","null"}`, else stringified — so `--data-eq stop_event_anchor_fallback_used=true` matches the boolean `True` (covered by a test).
It's threaded consistently through `build_event_report` (into `has_non_cycle_filter`, the `query_matches` conjunction, and the `_filter_report` filters echo), mirroring the existing `event_type`/`reason_code`/`tag` filter pattern exactly.

### Value-safety — within the local-tool boundary
This is the operator-local inspection CLI (`main` prints `json.dumps(report)` to stdout), not a shareable smoke/perf surface, so the value-safe-projection rule that governs those reports doesn't bind it — the tool already exposes raw event `data` by design, gated on `--include-data`. Two properties keep it clean regardless:
- **Filtering is orthogonal to output.** `_data_filters_match` reads `live_event["data"]` only to decide *inclusion*; it never injects field values into the output. You can filter by a data field and still get compact records *without* `--include-data` — the data is not forced into view. (Tests exercise both: the filter works, and data appears only when `--include-data` is passed.)
- **The `filters["data_eq"]` echo is the operator's own typed predicate** (key + expected values), shown back in their own local result — consistent with the pre-existing echoes of `event_types`/`reason_codes`/`statuses`/`tags`. Not a new exposure of event data.

### Robustness / UX
The `build_event_report(...)` call in the CLI is now wrapped in `try/except ValueError → parser.error(str(exc))`, so a malformed `--data-eq` yields a clean argparse error (exit 2) instead of a traceback — verified by a CLI test asserting `SystemExit` code 2 with `"key=value"` on stderr.

### Tests
Ran `tests/test_live_event_query.py` in a clean /tmp worktree at the PR head: **30 passed, 0 failed, 0 errors**. Coverage spans unit (`build_event_report`) and CLI (`main`) levels, happy path (AND-of-keys + bool normalization) and error path (malformed → `ValueError` / `SystemExit` 2). Fixtures use the #870 anchor events — the natural use case.

### Scope note (non-blocking)
By design (help text: "top-level event data field") the filter matches only top-level `data` keys with scalar string-equality (`_data_value_matches` stringifies; a nested dict/list would only match its repr). That's the right scope for this feature — just noting that nested-field or numeric-range predicates are intentionally out of scope, in case a follow-up wants them.

Net: a clean, well-tested, read-only operator-tooling addition that respects the local-tool/value-safety boundary and directly supports the flat-finalization-fallback diagnostics. Approving.

enarjord (commented):
**PR:** #873 — Filter live event queries by data fields
**Head SHA:** `dc30a4c39bef137babe9400ef13ea43d49440188` (was `fc63a3ae` when I approved)
**Base:** v8 (merge-base `c45537c1` = current tip)
**Verdict:** ✅ **Approve carries** — benign rebase, no code change

The head moved `fc63a3ae` → `dc30a4c39b`, but this is a **pure rebase** onto the new v8 tip (picking up #872's merge), not a content change. Verified: the diff of #873's own three files (`src/live/event_query.py`, `src/tools/live_event_query.py`, `tests/test_live_event_query.py`) between my approved head and this head is **empty** — byte-identical. The only difference between the two heads is `src/live/smoke_report.py` + `tests/test_live_smoke_report.py`, which is #872 (already merged and reviewed), now part of the rebased base rather than this PR's delta. #873's own-delta vs its merge-base `c45537c1` is the same `event_query.py +60 / live_event_query.py +64−26 / test +122` I reviewed.

My prior review stands unchanged: read-only local `live-event-query` `--data-eq` filter, value-safety boundary intact (filtering orthogonal to `--include-data`), correct AND-of-ORs + bool/None normalization + malformed→`parser.error` (exit 2), 30 passed. No re-test needed (code identical). Approving the rebased head.

PR #876, opened 2026-06-29
Record query-scope deploy evidence

enarjord:
## Summary
- record PR #875 merge/deploy/smoke evidence in the live logging progress ledger
- add the observed slow rotated live-event-query follow-up to the live ops backlog

## Scope
- Docs/progress only. No code, tests, event producers, exchange calls, cache mutation, readiness gates, order/risk logic, console routing, or trading behavior changed.

## Evidence recorded
- PR #875 merged at fda17fad and deployed to VPS5 without restarting bots
- 5-minute VPS5 brief smoke was ok=true, hard_failures=0, matched_expected=5, remote_calls.failed=0, account_critical_remote_calls.failed=0
- focused Gate.io/ZEC rotated event query with exchange/user scoping still ran for about two minutes before manual interruption; leftover probe process was cleaned up

## Validation
- git diff --check

PR #877, opened 2026-06-29
Prune old event segments for windowed queries

enarjord:
## Summary
- skip event segment files whose filesystem mtime is strictly before a requested since_ms/recent-minutes window
- report files_skipped_before_window in event_window so skipped files are observable
- add regression coverage proving old rotated files are skipped while recent/current files are still scanned

## Scope
- Read-only live-event-query tooling only.
- No event producers, exchange calls, cache mutation, readiness gates, order/risk logic, console routing, or trading behavior changed.

## Rationale
A VPS5 Gate.io/ZEC query with --exchange/--user, --include-rotated, and --recent-minutes 180 still ran for roughly two minutes because old rotated files were opened and row-filtered. This keeps row-level filtering authoritative for possibly relevant files, but avoids opening segments whose mtime proves they cannot contain in-window rows.

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/live/event_query.py tests/test_live_event_query.py
- git diff --check
- rg -n "except Exception|return_exceptions=True|\.get\([^\n]*,\s*(0|0\.0|None|False|\{\}|\[\])\)" src/live/event_query.py tests/test_live_event_query.py

### DIFF
--- a/src/live/event_query.py
+++ b/src/live/event_query.py
@@ -107,6 +107,13 @@ def _open_text(path: Path):
     return open(path, "r", encoding="utf-8")
 
 
+def _file_mtime_ms(path: Path) -> int | None:
+    try:
+        return int(path.stat().st_mtime * 1000)
+    except OSError:
+        return None
+
+
 def _live_event_payload(row: dict[str, Any]) -> dict[str, Any] | None:
     payload = row.get("payload")
     if not isinstance(payload, dict):
@@ -987,6 +994,7 @@ def build_event_report(
     window_skipped_before = 0
     window_skipped_after = 0
     window_invalid_ts = 0
+    files_skipped_before_window = 0
     issues: list[EventIssue] = []
     try:
         files = discover_event_files(
@@ -1010,6 +1018,15 @@ def build_event_report(
                 "no event NDJSON files found",
             )
         )
+    if since_filter is not None and files:
+        window_files: list[Path] = []
+        for event_file in files:
+            file_mtime_ms = _file_mtime_ms(event_file)
+            if file_mtime_ms is not None and file_mtime_ms < since_filter:
+                files_skipped_before_window += 1
+                continue
+            window_files.append(event_file)
+        files = window_files
 
     event_type_counts: Counter[str] = Counter()
     cycle_counts: Counter[str] = Counter()
@@ -1429,5 +1446,6 @@ def build_event_report(
             "events_skipped_before": window_skipped_before,
             "events_skipped_after": window_skipped_after,
             "invalid_window_ts": window_invalid_ts,
+            "files_skipped_before_window": files_skipped_before_window,
         }
     return report
--- a/tests/test_live_event_query.py
+++ b/tests/test_live_event_query.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import gzip
 import json
+import os
 
 import pytest
 
@@ -893,6 +894,7 @@ def test_event_query_filters_by_inclusive_time_window(tmp_path):
         "events_skipped_before": 1,
         "events_skipped_after": 1,
         "invalid_window_ts": 1,
+        "files_skipped_before_window": 0,
     }
     assert report["event_types"] == {
         "cycle.completed": 1,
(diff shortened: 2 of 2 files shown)


enarjord (commented):
**PR:** #877 — Prune old event segments for windowed queries
**Head SHA:** `47a1ec719acdc3fdb43f20b2b300e1588b864562`
**Base:** v8 (merge-base `ad3dc1ad` = current tip — branch is right at HEAD)
**Scope:** local event-query tooling — `src/live/event_query.py` (+18), `tests/test_live_event_query.py` (+71). Read-only NDJSON inspection; no live/trading path.
**Verdict:** ✅ **Approve** — no blockers (1 non-blocking assumption note)

This directly closes the backlog item #876 filed: it skips event NDJSON segments whose file mtime is entirely before the query's `since` bound, so time-windowed/`--recent-minutes` incident queries don't open every rotated segment (the ~2-minute Gate.io/ZEC case). Reviewed with the correctness focus this needs — a false prune would silently drop in-window events.

### Correctness — the prune is conservative and sound
```python
if since_filter is not None and files:
    for event_file in files:
        file_mtime_ms = _file_mtime_ms(event_file)
        if file_mtime_ms is not None and file_mtime_ms < since_filter:
            files_skipped_before_window += 1; continue
        window_files.append(event_file)
```
- **Only the `since` side is pruned, and that's the only side mtime can soundly support.** A segment's mtime is the time of its last append, i.e. an upper bound on the event timestamps it contains (events are appended in chronological order). So `mtime < since` ⟹ every event in the file predates the window ⟹ safe to skip. The `until` upper bound is correctly *not* used for pruning (that would need the file's earliest event time, which mtime can't provide) — good asymmetry.
- **Boundary-correct:** strict `<` means a segment with `mtime == since` (or spanning the boundary, mtime ≥ since) is kept, so a rotated segment straddling the window start is retained and its in-window events surface via the existing event-level filter.
- **Fail-open:** unreadable mtime (`OSError → None`) keeps the file rather than skipping it.
- **Opt-in / behavior-preserving:** the prune only runs when `since_filter is not None`; unbounded queries scan everything exactly as before. The active `current.ndjson` (mtime ≈ now) is never pruned for a recent window.
- **Belt-and-suspenders:** this is a file-open optimization layered *on top of* the authoritative per-event window filter (`events_skipped_before/after`), which remains the source of truth for any file that is scanned.

### Observability / value-safe
Adds a `files_skipped_before_window` counter to the `event_window` report block, alongside the existing `events_skipped_before/after`/`invalid_window_ts` — a count, value-safe, mirrors the established pattern. `_file_mtime_ms` is a clean `stat`-with-`OSError→None` helper. No trading path, producers, or new event data.

### Tests
Ran `tests/test_live_event_query.py` in a clean /tmp worktree at the PR head: **35 passed, 0 failed, 0 errors** (+1). The new `test_event_query_prunes_rotated_files_before_time_window_by_mtime` sets explicit mtimes via `os.utime` on old/recent/current segments and asserts `files_scanned == 2`, `files_skipped_before_window == 1`, `events_considered == 2`, and that exactly `cy_recent`+`cy_current` return in order (the pre-window `cy_old` segment pruned). The existing inclusive-window test is updated with `files_skipped_before_window: 0`.

### Non-blocking note — the mtime-as-upper-bound assumption
The prune trusts that a segment's mtime is ≥ the newest event timestamp it contains. That holds for normal append-only monitor writes (and the active file is always kept), so for real operation this is correct. The one theoretical false-prune is a segment whose mtime is anomalously *older* than its content — e.g. a backup/restore or `rsync` that preserves a stale mtime while the file holds newer events. Given this is the read-only diagnostic tool (not the trading path) and such mtimes don't arise from the monitor writer, it's an acceptable optimization; worth stating the assumption in case a future workflow ingests externally-copied segments. (If ever a concern, gating the skip on `name != "current.ndjson"` is already implicit via mtime, and a first-line-timestamp peek could harden it — not needed now.)

Net: a correct, conservative, well-tested windowing optimization that resolves the #876 rotated-scan-perf backlog item without weakening the authoritative event-level filtering. Approving.

vigilpbclaw-hash:
PR: #877
Head SHA: 47a1ec719acdc3fdb43f20b2b300e1588b864562
Base SHA/branch: ad3dc1ad19d57ccf9598a9eccacacaa65d6fda7c/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: None.
- Important: None.
- Minor: The pruning relies on the monitor segment filesystem `mtime` being at least as new as the newest row timestamp in that file. That matches normal append/rotate behavior and the helper fails open on stat errors, but externally restored/copied segments with stale preserved mtimes could be skipped by a `since_ms` query. I do not consider this blocking for the read-only diagnostic query path.

Intent/architecture pass:
The PR is trying to make `live-event-query --include-rotated --recent-minutes/--since-ms` usable on large rotated histories without weakening row-level filtering. Adding a conservative pre-scan file prune in `build_event_report` fits the existing abstraction: `discover_event_files` still handles path/scope discovery, while the query layer applies the time-window optimization and reports it via `event_window.files_skipped_before_window`. It is observability/tooling-only; I found no changes to producers, exchange calls, readiness gates, order/risk logic, console routing, or live trading behavior.

Evidence reviewed:
- `src/live/event_query.py`: `_file_mtime_ms()` returns `None` on `OSError`, so unreadable metadata keeps the file; pruning is only enabled when `since_ms` is present and uses strict `mtime_ms < since_ms`, preserving boundary/spanning files for authoritative per-row filtering.
- `tests/test_live_event_query.py`: new regression covers old rotated segment skipped while recent rotated/current segments are scanned and counted.

Validation run:
- `git diff --check refs/remotes/origin/v8...HEAD` — passed.
- `PYTHONPATH=src python3 -m pytest tests/test_live_event_query.py -q` — 35 passed (one pre-existing pytest config warning about `asyncio_mode`).
- `PYTHONPATH=src python3 -m compileall -q src/live/event_query.py tests/test_live_event_query.py` — passed.
- Additional targeted smoke for an old gz segment plus current segment with `since_ms` pruning — passed.
- Dry-run merge of PR head into current `origin/v8` — automatic merge succeeded.

PR #894, opened 2026-06-30
Emit exchange config refresh events

enarjord:
Adds structured live-event visibility for hourly exchange config/markets refresh results without changing maintenance behavior.\n\nScope:\n- new off-console/text event type: exchange.config_refresh\n- best-effort emitter with sanitized error payloads\n- hourly maintainance init_markets wrapper emits succeeded/failed events and re-raises original errors\n- registry docs and focused tests\n\nValidation:\n- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_exchange_config_refresh_event.py tests/test_live_event_bus.py tests/test_live_event_registry_docs.py -> 41 passed\n- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/event_bus.py src/live/event_emitters.py src/passivbot.py tests/test_exchange_config_refresh_event.py\n- git diff --check

PR #868, opened 2026-06-29
Surface HSL flat finalizations in live reports

enarjord:
## Summary
- include hsl.red_finalized_without_order in live smoke risk_events
- include the same event in live performance risk_activity
- keep projected fields bounded to provenance/ratio fields and extend value-safety tests

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_smoke_report.py tests/test_live_performance_report.py
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m compileall src/live/smoke_report.py src/live/performance_report.py tests/test_live_smoke_report.py tests/test_live_performance_report.py
- git diff --check

PR #896, opened 2026-06-30
Summarize exchange config refresh health

enarjord:
Adds a read-only live-smoke-report projection for the structured exchange.config_refresh events from PR #894.\n\nScope:\n- full report: exchange_config_refresh_health groups/totals\n- summary: limited exchange_config_refresh_health groups\n- brief: exchange_config_refresh top-level counters\n- excludes raw free-text data.error from smoke projections; keeps labels/error_type/timing only\n- no smoke verdict or text-log classification changes\n\nValidation:\n- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_smoke_report.py -> 52 passed\n- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/smoke_report.py tests/test_live_smoke_report.py\n- git diff --check

### DIFF
--- a/src/live/smoke_report.py
+++ b/src/live/smoke_report.py
@@ -73,6 +73,7 @@ EMA_READINESS_GROUP_LIMIT = 20
 STAGED_READINESS_GROUP_LIMIT = 20
 EVENT_PIPELINE_HEALTH_GROUP_LIMIT = 20
 HSL_REPLAY_HEALTH_GROUP_LIMIT = 20
+EXCHANGE_CONFIG_REFRESH_HEALTH_GROUP_LIMIT = 20
 RISK_EVENT_TYPES = {
     EventTypes.RISK_MODE_CHANGED,
     EventTypes.HSL_TRANSITION,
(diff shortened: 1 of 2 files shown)


vigilpbclaw-hash:
PR: #896
Head SHA: b8effba8d518594dc43cc3fe328208777400270c
Base SHA/branch: 5177f5e05ba811201a33a09cfe2f76859a4a396a/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: This PR adds a read-only smoke-report projection for `exchange.config_refresh` events created by the prior event-emission work. The implementation keeps the feature in the existing smoke-report event scan/projection pipeline, groups by bot/status/reason/operation/error_type, redacts the allowed string fields, and deliberately excludes raw `data.error` from report/summary/brief output. That matches the stated observability-only intent and does not introduce live-trading gates or behavior changes.

Evidence reviewed:
- Diff: `src/live/smoke_report.py`, `tests/test_live_smoke_report.py` against base `5177f5e05ba811201a33a09cfe2f76859a4a396a`.
- Checked existing event emitter payload shape in `src/live/event_emitters.py` and maintenance call sites in `src/passivbot.py` for compatibility.
- Validation run: `git diff --check origin/v8...HEAD` (passed).
- Validation run: `python3 -m py_compile src/live/smoke_report.py tests/test_live_smoke_report.py` (passed).
- Validation run: `python3 -m pytest tests/test_live_smoke_report.py::test_live_smoke_report_summarizes_exchange_config_refresh_health tests/test_live_smoke_report.py::test_live_smoke_report_brief_summary_projects_top_level_counters -q` (2 passed).
- Broader `python3 -m pytest tests/test_live_smoke_report.py -q` had 2 failures in HSL balance-override config-check tests; I reproduced the same two failures on the PR base SHA, so they appear unrelated to this PR/environment rather than introduced here.

enarjord (commented):
**PR:** #896 — Summarize exchange config refresh health
**Head SHA:** `b8effba8d518594dc43cc3fe328208777400270c`
**Base:** v8 (merge-base `5177f5e0` = current tip — branch is right at HEAD)
**Scope:** read-only smoke-report projection — `src/live/smoke_report.py` (+181), `tests/test_live_smoke_report.py` (+91). No live path, no verdict/text-log change.
**Verdict:** ✅ **Approve** — no blockers (1 forward-looking note for the eventual verdict-reclassification slice)

This is the **smoke/report projection step of backlog #18** — surfacing the structured `exchange.config_refresh` events from #894 in the smoke report. It's the highest-scrutiny class of change (the #18 health-gate watch), so I reviewed it against the full SMOKE-REPORT value-safety contract and the #18 guardrails. It is correctly the *conservative, additive* slice: it makes the refresh outcome **visible** without touching the pass/fail gate.

### Value-safe — raw `data.error` free-text excluded from every surface (the load-bearing property)
`_exchange_config_refresh_group` builds `latest_data` from a **fixed whitelist only** — `context`, `operation`, `error_type` (each `_redact_log_text(..., max_len=120)`) plus `elapsed_ms`/`started_ms` (`_non_negative_int`). It never reads `payload["error"]`. So the bounded sanitized free-text error (which #894's emitter carries in off-console NDJSON) is dropped at the projection boundary — envelope labels + `error_type` + timing only, exactly as #894's forward-looking note required.
- **Full report** `exchange_config_refresh_health.groups` are `compact_groups` that additionally strip `latest_path`/`latest_line`/`latest_seq` — so even the local/full surface carries no path for this block.
- **Summary** (shareable) via `_summary_limited_groups` passes through only those pre-stripped compact groups + fixed count/label keys (incl. the new `statuses` status→count map). No path, no error.
- **Brief** (shareable) carries only `total`/`bots`/`succeeded`/`failed`/`failure_pct`/`failed_bots`/`event_types` — counts + a label→count dict.

The new test **proves** this, not just asserts the happy path: it injects `error: "binanceusdm apiKey=supersecret code=-4084"` into a failed event, then asserts `groups[0].latest_data` equals exactly the 5 whitelisted keys (no `error`) and that `json.dumps(health)` contains neither `supersecret`, `apiKey`, nor `code=-4084`. `error_type: "ExchangeError"` (a class-name label) is correctly kept as the classification signal. No equity/drawdown/pnl/balance/price anywhere.

### No verdict / text-log classification change — correct #18 sequencing
The block is **purely additive**: it's emitted into the report dict and the summary/brief projections, but nothing consumes it for `ok`, `issues`, or any verdict aggregation (verified — all references are the builder + the three projections). So the brief smoke still goes `ok=false` from the pre-existing Binance `-4084` text-log hard match. That's the intended order — make criticality *visible* via structured health first; the actual down-classification of the verdict is a later, separately-reviewed slice. The PR body states this explicitly ("no smoke verdict or text-log classification changes"). Good.

### Shared-whitelist lesson applied
`EXCHANGE_CONFIG_REFRESH` is **not** added to `RISK_EVENT_TYPES`; it gets its own dedicated `if event_type == EventTypes.EXCHANGE_CONFIG_REFRESH` branch and a fixed-key builder. So the shared `_risk_event_group` whitelist (which would otherwise spread `data` wholesale and leak drawdown_*/red_threshold/error) is not in play here — the leak class that bit the #868 drawdown-pop can't occur because nothing spreads `data`.

### Bounded
`EXCHANGE_CONFIG_REFRESH_HEALTH_GROUP_LIMIT = 20`, groups truncated to 20 with a `groups_truncated` flag; `event_types`/`statuses` are `Counter.most_common()` (bounded by distinct types/statuses). Grouping key = (bot, status, reason_code, operation, error_type) with count rollup and latest-by-position update via `_sort_event_position_key`. Mirrors the established sibling health blocks (ema/staged/event-pipeline/hsl-replay).

### Tests
Ran `tests/test_live_smoke_report.py` in a clean /tmp worktree at the PR head (`b8effba8`): **52 passed, 0 failed, 0 errors** — matches the PR's reported count.

### Forward-looking note (for the eventual verdict-reclassification slice, non-blocking here)
This PR deliberately does **not** flip the smoke verdict. When a follow-up actually uses `exchange_config_refresh_health` to down-classify the `-4084` maintenance traceback (so `ok` can pass despite it), that is the slice that needs MAX care: it must only down-classify **recurring NON-critical maintenance refresh** (e.g. a bounded recurrence of `status="failed"` from the hourly `init_markets` hedge-mode refresh), and must **never** down-classify a trading-critical startup-config or order-construction failure — i.e. preserve fail-loop/fail-safe-on-missing-proof. The structured `status`/`reason_code`/`error_type`/recurrence signals this PR now exposes are the right inputs for that gate; this PR correctly leaves the gate itself untouched.

Net: a clean, contract-complete, well-tested projection that makes the exchange config-refresh health explicit and shareable while excluding the raw error free-text and leaving the smoke verdict unchanged — exactly the right next step for #18. Approving.

PR #918, opened 2026-06-30
Add incident bundle query scope filters

enarjord:
## Summary
- expose common live-event-query scope filters on live-incident-bundle: level, exchange, user, bot-id, remote-call-group-id, side, source, component, tag, and data-eq
- pass those filters to both event_report.json and problem_event_report.json, and record them in manifest filters
- keep smoke/process/log behavior unchanged and fold #917 deploy evidence into the progress/backlog docs

## Scope
Read-only operator tooling only. No event producers, exchange calls, cache mutation, readiness gates, console routing, monitor writes, order/risk logic, or trading behavior changed.

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/incident_bundle.py src/tools/live_incident_bundle.py
- git diff --check
- silent-handling scan on touched report/test files; remaining hits are pre-existing non-trading report summary defaults in incident_bundle.py

PR #916, opened 2026-06-30
Filter live event query problem events

enarjord:
## Summary

- add `live-event-query --problem-events` and `--hard-problem-events`
- reuse the same structured problem-event predicate as `live-smoke-report` attention/hard-failure accounting, including the HSL shutdown-cancel exception
- keep filters composable with existing event type, bot, symbol, status, level, tag, data, and time-window scopes
- fold PR #915 degraded-gate merge and VPS5 smoke evidence into the progress/backlog docs

## Scope

Read-only query/report tooling only. No event producers, exchange calls, live execution, report verdict changes, console routing, order/risk logic, or trading behavior.

## Validation

- `PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py`
- `/Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/event_query.py src/tools/live_event_query.py tests/test_live_event_query.py`
- `git diff --check`
- touched-file silent-handling scan: no matches

PR #907, opened 2026-06-30
Prune live event queries by path bot id

enarjord:
## Summary
- prune live-event-query file discovery for path-shaped bot ids like exchange/user
- preserve existing behavior for opaque event-level bot ids like bot_1 by not path-pruning them
- record PR #906 deploy and VPS smoke evidence in the logging progress/backlog docs

## Validation
- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_event_query.py
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/event_query.py tests/test_live_event_query.py
- git diff --check
- touched-file silent-handling scan: no matches

## VPS evidence from prior merged slice
- VPS5 pulled v8 to b7b34758 without bot restart; all five configured bots left running
- focused Binance current-segment query with --source/--component returned ok=true and files_scanned=1
- focused Binance current-segment query with --side buy,sell returned ok=true and matched one event
- bounded 2-minute brief smoke returned ok=true, hard_failures=0, matched_expected=5, remote_calls.failed=0, account_critical_remote_calls.failed=0, repo dirty=false

Observability/query tooling plus progress docs only; no event producers, exchange calls, live execution, or trading behavior changed.

PR #919, opened 2026-06-30
Scope incident bundle time-window events

enarjord:
## Summary\n- applies incident-bundle query scope filters to time_window_report.json\n- keeps timeline.txt and matched event-segment selection scoped to the same focused bundle rows\n- preserves existing cycle-id behavior for surrounding time-window context: cycle-focused bundles still include other matching events in the requested time window\n- updates logging-overhaul progress/backlog ledger for PR #918 and this follow-up slice\n\n## Behavior boundary\nRead-only incident-bundle/report tooling only. No event producers, exchange calls, live execution, report verdict changes, console routing, monitor writes, cache mutation, readiness gates, order/risk logic, or trading behavior changed.\n\n## Validation\n- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py\n- PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/incident_bundle.py src/live/event_query.py src/tools/live_incident_bundle.py\n- git diff --check\n\n## Notes for reviewers\nThis follows the VPS5 finding after PR #918: focused OKX bundles correctly scoped event_report/problem_event_report, but time_window_report/timeline/segment selection still scanned broader root context.


enarjord:
Reviewer: passivbot-cursor-agent
PR: #919
Head SHA: ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001 (v8)
Scope reviewed: full PR
Verdict: approve

**Findings:**

- No blockers.
- **Important (architecture):** `event_matches_query_filters` (new, `src/live/event_query.py`) is a byte-for-byte reimplementation of the per-event filter-matching logic that already lives inline inside `build_event_report`'s main scan loop (`event_type_matches`/`level_matches`/.../`problem_matches` block, currently around `src/live/event_query.py:1560-1633`, including the `hard_problem_event_filter` > `problem_event_filter` precedence and the `path_scope` exchange/user fallback). `build_event_report` itself was **not** refactored to call the new helper — it still has its own copy. That means there are now two independent sources of truth for "does this event match these filters," used respectively by `event_report`/`problem_event_report` (inline loop) and `time_window_report` (new helper via `_build_time_window_report`). A future filter addition or semantics tweak applied to one path and not the other would silently make `event_report` and `time_window_report` disagree on what counts as "matching" inside the same incident bundle — exactly the kind of cross-report scoping inconsistency this PR is trying to eliminate (per the VPS5 finding cited in the PR description). Recommend a fast follow-up that refactors the inline loop in `build_event_report` to call `event_matches_query_filters` instead of keeping both, so there is a single predicate implementation.
- The fix itself (scoping `time_window_report`/`timeline.txt`/event-segment selection to the same filters as `event_report`/`problem_event_report`) is correct and directly addresses the regression: `_build_time_window_report` now builds `query_filters` via `build_event_query_filters`, filters discovered files via `discover_event_files(..., exchange=, user=, bot_id=)` (reusing existing pruning kwargs, not new logic), and skips non-matching rows via `event_matches_query_filters` before counting them as `matched_events`.
- Cycle-focused bundles still surface the full surrounding-time-window context as before — `_build_time_window_report`'s call site explicitly forwards all the new filters with a comment noting this preserved behavior, and it's exercised by the extended `test_live_incident_bundle_cli_filters_event_reports_by_query_scopes` (now asserting `timeline.txt`, `event_segments_manifest.json`, and `time_window_report.json` are all scoped to the same single OKX event, while the binance event with a different scope is excluded).
- `build_event_query_filters` raises `ValueError("since_ms must be <= until_ms")` redundantly with the existing CLI-level `since_ms <= until_ms` check in `live_incident_bundle.py`'s `main()` — harmless belt-and-suspenders, not a concern.

**Architecture / design fit:** mostly sound — extracting `build_event_query_filters`/`event_query_filter_report`/`event_matches_query_filters` as reusable building blocks is the right direction so other report sections can apply the same scope filters. The gap is that the extraction wasn't also wired back into `build_event_report`'s existing loop, leaving a duplicate implementation behind. Important enough to flag, not blocking given current behavior is correct and tested.

**Tests run:**
- `PYTHONPATH=src venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py -q` → all passed.
- Diffed `src/live/event_query.py` and `src/live/incident_bundle.py` against base to confirm `event_matches_query_filters` is not called from `build_event_report`'s scan loop, and confirmed `discover_event_files(exchange=, user=, bot_id=)` kwargs are pre-existing (not new pruning logic).

Read-only incident-bundle/report tooling only — no event producers, exchange calls, live execution, console routing, monitor writes, cache mutation, readiness gates, or order/risk logic touched, consistent with the PR description.

enarjord (commented):
**PR:** #919 — Scope incident bundle time-window events
**Head SHA:** `ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e`
**Base:** v8 (merge-base `946d0757` = current tip — branch is right at HEAD)
**Scope:** event-query + incident-bundle library + docs — `src/live/event_query.py` (+175), `src/live/incident_bundle.py` (+82/−12), `tests/test_live_incident_bundle.py` (+24), docs (+51/−3). No live path, no producers, no verdict change.
**Verdict:** ✅ **Approve** — no blockers (1 Minor)

Completes the #918 scoping work for the bundle's **time-window** report: it extracts the event-query filter machinery into three reusable helpers and uses them so the time-window report honors the same scope filters as the event/problem reports.

### What it does
- New `build_event_query_filters(...)` normalizes the full filter set (via the existing `_normalize_filter_values` / `_normalize_data_eq_filters`) and keeps the `since_ms <= until_ms` `ValueError` guard.
- New `event_query_filter_report(filters)` projects the normalized filters into the report `filters` echo (wraps `_filter_report`).
- New `event_matches_query_filters(path, row, live_event, filters)` is the per-event matcher.
- `_build_time_window_report` now builds `query_filters` from these, prunes discovery via `discover_event_files(..., exchange=, user=, bot_id=)` (the #907 path pruning), and `continue`s on any event that fails `event_matches_query_filters`. `build_live_incident_bundle` threads the same scope args into it.

### Behavior-preserving default + thoughtful cycle handling
- All scope filters default `None`, so `event_matches_query_filters` returns `True` for every event and the default time-window report is byte-for-byte unchanged (all events in `[since, until]`). Only an explicitly-scoped bundle narrows the window.
- `cycle_id` is deliberately passed as `None` to the window filters (with a code comment): a cycle-focused bundle still shows the surrounding-window context rather than collapsing to just that cycle. Good call — it keeps the window report's "what else happened around the incident" value while still letting symbol/exchange/etc. scope it.

### Read-only / value-safe
The time-window report stays part of the local redacted-evidence archive. The new `filters` echo records operator filter **criteria** (labels + `data_eq` key=value + `since_ms`/`until_ms` timestamps) — no event-data magnitudes; the captured window events are the same compact records as before, just scoped. The bundle's shareable `time_window` summary block (counts) is unchanged. No producers/exchange/verdict/trading changes.

### Minor — the new matcher duplicates `build_event_report`'s inline matcher (drift risk)
`event_matches_query_filters` / `build_event_query_filters` / `event_query_filter_report` are a faithful extraction of the matching logic, **but `build_event_report` was not refactored to use them** — it still carries its own inline `*_matches` chain + filter normalization + `_filter_report` call. So there are now two copies of the event-matching predicate. I compared them and they're equivalent today (same filter set, same `_filter_matches`/tag-intersection/`_data_filters_match` logic, same `is_problem_event`/`is_hard_problem_event` shared predicate), so this isn't a current bug. But the whole point of this slice is that the window report matches events *the same way the event report does* — if `build_event_report`'s inline matcher later changes (a new filter, a fallback tweak) and the extracted copy isn't updated in lockstep, the time-window and event reports would silently diverge.

The fix is right there: since the three helpers are already extracted, route `build_event_report` through them too (build its `query_filters` via `build_event_query_filters`, match via `event_matches_query_filters`, echo via `event_query_filter_report`) so there's a single source of truth — the same de-dup you did for the problem-event predicate in #916 (`live/problem_events`). Not blocking; flagging so the second copy doesn't outlive this PR.

### Tests
Ran `tests/test_live_incident_bundle.py` + `tests/test_live_event_query.py` in a clean /tmp worktree at the PR head (`ecba6053`): **62 passed, 0 failed, 0 errors** (`event_query.py` + `incident_bundle.py` `py_compile`-clean). The extended `..._cli_filters_event_reports_by_query_scopes` now reads `time_window_report.json` and asserts its scoped `filters` echo (matching the event/problem reports + `since_ms`), `matched_events == 1`, and `events[0].exchange == "okx"` — locking in that the window scopes consistently with the other reports. (A direct unit test of `event_matches_query_filters` vs `build_event_report` agreement on a shared fixture would be the natural guard for the Minor above.)

### Docs
Head-pointer bumped `0f32aeff`→`946d0757` (#918) = current tip; folds #918's deploy evidence and records #919 as pending. Faithful, value-safe, no overclaim; coin-HSL replay #777/#802 untouched.

Net: a clean completion of bundle scoping for the time-window report, behavior-preserving by default, with a sensible cycle-context carve-out and consistent file pruning. Approving; the one Minor is to fold `build_event_report` onto the newly-extracted shared matcher so the two copies can't drift.

vigilpbclaw-hash:
PR: #919
Head SHA: ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001/v8
Scope reviewed: full PR
Verdict: needs fix

- Blocker: none.
- Important: `src/live/event_query.py` now has `event_matches_query_filters()` as the predicate used by `_build_time_window_report()`, but `build_event_report()` still maintains a separate in-line copy of the same query predicate around its scan loop (`event_type_matches`, `level_matches`, `cycle_matches`, path-derived exchange/user fallback, bot-id fallback, tag/data/problemevent precedence, etc.). The PR intent is to make incident bundle report sections apply the same query scope; leaving two implementations means the next filter/semantics change can silently reintroduce disagreement between `event_report`/`problem_event_report` and `time_window_report`/timeline/segments. Please make `build_event_report()` call the extracted predicate (or otherwise centralize the predicate) so there is one source of truth for query matching.
- Minor: none.

Intent/architecture pass: this is a read-only incident-bundle/reporting PR intended to scope time-window reports, timeline output, and event-segment selection to the same filters as focused event reports. The behavioral fix is directionally right and uses existing discovery pruning for exchange/user/bot paths. The remaining design gap is the duplicated matching predicate, which is especially risky because this area is explicitly about keeping cross-report data contracts aligned.

Validation run:
- Dry-run merge of `origin/pr-919` into current `origin/v8` in a separate `/tmp` worktree — merged cleanly.
- `python3 -m py_compile src/live/incident_bundle.py src/live/event_query.py` — passed.
- `PYTHONPATH=src python3 -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py -q` — 62 passed, 1 pytest config warning about `asyncio_mode`.

No event producers, exchange calls, live execution, order/risk logic, readiness gates, cache mutation, or monitor write paths appear to be changed.

enarjord:
Updated head to 9f1741834 to address the duplicate-predicate finding from Cursor/Hermes/Claude. build_event_report now builds normalized filters with build_event_query_filters() and uses event_matches_query_filters() for query matching, so event_report/problem_event_report and time_window_report share one predicate. Re-ran: PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py (62 passed), py_compile for touched modules, and git diff --check.

vigilpbclaw-hash:
PR: #919
Head SHA: 9f17418345febdd743cf160b457d52d951bfd51f
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001/v8
Scope reviewed: delta ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e..9f17418345febdd743cf160b457d52d951bfd51f
Verdict: approve

Previous Hermes review for head `ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e` is superseded by this new head SHA.

- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: the delta addresses the prior duplicated-predicate design concern by making `build_event_report()` construct normalized filters with `build_event_query_filters()` and evaluate rows through `event_matches_query_filters()`. That aligns the event report/problem-event report path with the time-window/timeline/segment selection path and is the right practical architecture for this read-only incident-bundle scoping fix: one shared query predicate for the cross-report data contract. I did not find evidence of trading behavior, event producers, exchange calls, live execution gates, order/risk logic, monitor writes, or cache mutation changes.

Validation run:
- Dry-run merge of `origin/pr-919` into current `origin/v8` in a separate `/tmp` worktree — merged cleanly.
- `python3 -m py_compile src/live/incident_bundle.py src/live/event_query.py` — passed.
- `PYTHONPATH=src python3 -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py -q` — passed: 62 tests passed, with one existing pytest config warning about unknown `asyncio_mode`.

enarjord (commented):
**PR:** #919 — Scope incident bundle time-window events
**Head SHA:** `9f17418345febdd743cf160b457d52d951bfd51f` (amended from `ecba6053`)
**Base:** v8 (merge-base `946d0757` = current tip — branch is right at HEAD)
**Scope (amendment):** `src/live/event_query.py` only (+56/−112, net −56). Tests/docs/incident_bundle unchanged from the prior head.
**Verdict:** ✅ **Approve** — no blockers (prior Minor resolved; 1 new low-severity Minor)

The amendment implements the Minor from my prior review: `build_event_report` now routes through the newly-extracted `build_event_query_filters` + `event_matches_query_filters`, deleting its ~110-line inline matcher. There's now a single matcher shared by the event report and the incident-bundle time-window report — the drift risk is gone. Since this changes the matcher that drives event-query, the problem report, and the incident event report, I fresh-reviewed the full inter-head delta and ran the broad suite.

### Single source of truth — faithful de-dup
- `build_event_report` builds `query_filters = build_event_query_filters(...)`, then reads each `*_filter` from `query_filters[...]` (so the `has_non_cycle_filter`/`has_query_filter` gating is computed from the same sets as before), and the `since`/`until` guard now comes from `build_event_query_filters` (same `ValueError("since_ms must be <= until_ms")`).
- The whole inline `*_matches` chain + `query_matches = (… and …)` is replaced by one `event_matches_query_filters(path=…, row=…, live_event=…, filters=query_filters)` call.
- The per-event record derivations live in the matcher now and are identical to the deleted inline ones (`record_event_type = live_event.get("event_type") or row.get("kind")`, exchange/user with `_monitor_path_exchange_user` fallback, `ids.get("bot_id") or _monitor_path_bot_id(path)`, `_event_tags`, `_data_filters_match`, shared `is_problem_event`/`is_hard_problem_event`). All non-cycle scalar filters go through `_filter_matches`, which does `str(value) in filter_values`, so int/str coercion is preserved exactly. `record_cycle_id`/`cycle_counts` bookkeeping stays in `build_event_report` (only the *match* moved).

### New Minor — cycle_id compare loses the old str-normalization on the event side
This is the one behavioral asymmetry the de-dup introduces. The pre-refactor inline matcher compared `str(record_cycle_id) == str(cycle_id)` — i.e. it str-normalized **both** the event's cycle_id (`record_cycle_id = str(ids.get("cycle_id"))`) and the filter. The shared matcher instead does:

```python
if cycle_id is not None and ids.get("cycle_id") != str(cycle_id):
    return False
```

— the filter side is str-normalized (`str(cycle_id)`), but the **event side `ids.get("cycle_id")` is not**. cycle_id is also the one field that bypasses `_filter_matches` (direct `==`), so unlike every other filter it doesn't get the `str(value)` coercion. For the contractual case this is a no-op: `LiveEvent.cycle_id` is typed `str | None` and the emitter writes `f"cy_{int(seq)}"`, so real events carry string cycle_ids and the match is identical. But it drops the old code's tolerance of a non-string `ids["cycle_id"]` (e.g. an int epoch like `planning_availability.cycle_id = int(snapshot.epoch)`): a `--cycle-id 5` filter would match `record_cycle_id == "5"` under the old code but fail `5 != "5"` under the new matcher. Low severity given the `str` type contract, but to keep exact parity with the matcher you replaced, normalize the event side too — e.g. `str(ids.get("cycle_id")) != str(cycle_id)` (the `None` case still correctly no-matches since `"None" != str(cycle_id)`). The None-missing and string cases are already equivalent; only the non-string case differs.

### Tests
Ran `tests/test_live_event_query.py` + `tests/test_live_smoke_report.py` + `tests/test_live_incident_bundle.py` in a clean /tmp worktree at the amended head (`9f174183`): **115 passed, 0 failed, 0 errors** (`event_query.py`/`incident_bundle.py`/`smoke_report.py` `py_compile`-clean). The full event-query suite passing confirms the inline→shared matcher swap preserves behavior for the (string) cycle_id and all `_filter_matches`-routed fields; the incident suite still asserts the time-window report scopes consistently. (A unit test with a non-string `ids["cycle_id"]` would pin the Minor above either way.)

Net: the amendment cleanly resolves my prior drift-risk Minor by giving the event report and the time-window report one matcher, and the refactor is behavior-preserving for real (string) cycle_ids. Re-approving the amended head; the single new Minor is to restore the event-side `str()` on the cycle_id compare so the shared matcher is byte-equivalent to the inline one it replaced.

enarjord:
Updated head to 270d91b49 for Claude's cycle_id parity minor. event_matches_query_filters now string-normalizes the event-side cycle_id before comparison, matching the old inline matcher. Added test_event_query_cycle_filter_matches_non_string_event_id. Re-ran: PYTHONPATH=src /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py (63 passed), py_compile for touched modules, git diff --check, and touched-file silent scan.

enarjord:
Reviewer: passivbot-cursor-agent
PR: #919
Head SHA: 9f17418345febdd743cf160b457d52d951bfd51f
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001 (v8)
Scope reviewed: delta ecba6053e80bdb769ab39f5bf20f5c31cd49aa6e..9f17418345febdd743cf160b457d52d951bfd51f (supersedes prior review at ecba6053)
Verdict: approve

**Findings:**

- No blockers or important issues remaining.
- This delta resolves the Important architecture concern from my previous review of `ecba6053`: `build_event_report`'s inline filter-matching block (`event_type_matches`/`level_matches`/.../`problem_matches`, the duplicate of `event_matches_query_filters`) has been deleted and replaced with a direct call to the shared `event_matches_query_filters(path=path, row=row, live_event=live_event, filters=query_filters)`. `build_event_report` now also builds its filters via `build_event_query_filters(cycle_id=cycle_id, ...)` instead of calling `_normalize_filter_values`/`_normalize_data_eq_filters` directly inline, so `cycle_id` is folded into the shared filter dict and matched inside `event_matches_query_filters` rather than via a separate `cycle_matches` inline check — verified this is handled correctly inside `event_matches_query_filters` (`cycle_id = filters.get("cycle_id"); if cycle_id is not None and ids.get("cycle_id") != str(cycle_id): return False`).
- Net diff is a 56/112 line reduction in `src/live/event_query.py` only — pure deduplication, no other files touched in this delta, consistent with a targeted follow-up.
- There is now a single source of truth for "does this event match these filters," used identically by `event_report`/`problem_event_report` (via `build_event_report`) and `time_window_report` (via `_build_time_window_report`), eliminating the drift risk I flagged.

**Architecture / design fit:** no remaining concerns — this is exactly the consolidation I suggested in the prior review.

**Tests run:**
- `PYTHONPATH=src venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py -q` → all passed (62 tests), confirming `build_event_report`'s refactored scan loop still produces identical filtering behavior.
- Diffed `src/live/event_query.py` between the two heads to confirm the inline duplicate block was removed (not left behind alongside the helper) and that no other files changed in this delta.

Read-only incident-bundle/report tooling only — no behavior change versus the previous head beyond internal deduplication; no event producers, exchange calls, live execution, console routing, monitor writes, cache mutation, readiness gates, or order/risk logic touched.

enarjord:
Reviewer: passivbot-cursor-agent
PR: #919
Head SHA: 270d91b49dbbb6fd27b740fb4357615bfdf15d77
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001 (v8)
Scope reviewed: delta 9f17418345febdd743cf160b457d52d951bfd51f..270d91b49dbbb6fd27b740fb4357615bfdf15d77 (supersedes prior review at 9f174183)
Verdict: approve

**Findings:**

- No blockers or important issues.
- This delta fixes a real regression introduced by the previous delta's deduplication: the original inline loop in `build_event_report` coerced `record_cycle_id` to `str(...)` before comparing against the `cycle_id` filter (`record_cycle_id = str(record_cycle_id)` then `record_cycle_id == str(cycle_id)`), but the new shared `event_matches_query_filters` compared `ids.get("cycle_id") != str(cycle_id)` without coercing the left side. For monitors that emit a non-string `cycle_id` (e.g. an int), this would have silently broken `--cycle-id` filtering for that event — a behavior regression versus pre-#919 for that specific case. Fixed with `str(ids.get("cycle_id")) != str(cycle_id)`.
- Good regression test added: `test_event_query_cycle_filter_matches_non_string_event_id` constructs an event with integer `cycle_id: 7` alongside one with string `cycle_id: "8"`, and confirms filtering on `cycle_id="7"` matches only the integer-keyed event. This is exactly the right kind of test to lock in the fix and guard the shared helper going forward.
- Diff is minimal and precisely scoped: one line in `src/live/event_query.py`, one new test. No unrelated changes.

**Architecture / design fit:** no concerns — this is a precise bug fix in the newly-consolidated shared predicate, which is exactly where such a fix belongs now that there is a single source of truth (per my prior reviews of this PR).

**Tests run:**
- `PYTHONPATH=src venv/bin/python -m pytest tests/test_live_incident_bundle.py tests/test_live_event_query.py -q` → all passed (63 tests, including the new regression test).
- Verified the one-line fix matches the original inline behavior (`str(record_cycle_id) == str(cycle_id)` semantics) that existed before the #919 consolidation.

Read-only incident-bundle/event-query tooling only — bug fix restores prior cycle-id filtering semantics for non-string ids; no order/risk logic touched.

vigilpbclaw-hash:
PR: #919
Head SHA: 270d91b49dbbb6fd27b740fb4357615bfdf15d77
Base SHA/branch: 946d0757f36c8c5994117bf26979e8d68105d001/v8
Scope reviewed: delta 9f17418345febdd743cf160b457d52d951bfd51f..270d91b49dbbb6fd27b740fb4357615bfdf15d77
Verdict: approve

Previous Hermes review for head `9f17418345febdd743cf160b457d52d951bfd51f` is superseded by this new head SHA.

- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this delta is a targeted follow-up to preserve exact pre-refactor cycle-id matching semantics in the newly shared event-query predicate. It changes `event_matches_query_filters()` to coerce the event-side `ids["cycle_id"]` with `str(...)` before comparing to the requested cycle id, matching the old inline `build_event_report()` behavior and keeping the single-source-of-truth architecture from the previous head. The added regression test covers a non-string event cycle id (`7`) filtered by string cycle id (`"7"`). This is the right place for the fix and does not expand scope beyond read-only incident-bundle/event-query reporting.

Validation run:
- Final duplicate guard checked comments/reviews immediately before posting: no Hermes Continuous Reviewer comment/review existed for head `270d91b49dbbb6fd27b740fb4357615bfdf15d77`.
- `git diff --check origin/v8...origin/pr-919` — passed.
- Dry-run merge of `origin/pr-919` into current `origin/v8` in a separate `/tmp` worktree — merged cleanly.
- `python3 -m py_compile src/live/event_query.py src/live/incident_bundle.py` — passed.
- `PYTHONPATH=src python3 -m pytest tests/test_live_event_query.py tests/test_live_incident_bundle.py --tb=short` — passed: 63 passed, 1 pytest config warning about unknown `asyncio_mode`.

No event producers, exchange calls, live execution, order/risk logic, readiness gates, cache mutation, or monitor write paths appear to be changed.

enarjord (commented):
**PR:** #919 — Scope incident bundle time-window events
**Head SHA:** `270d91b49dbbb6fd27b740fb4357615bfdf15d77` (amended from `9f174183`)
**Base:** v8 (merge-base `946d0757` = current tip — branch is right at HEAD)
**Scope (amendment):** `src/live/event_query.py` (1-line) + `tests/test_live_event_query.py` (+28). Nothing else changed from the prior head.
**Verdict:** ✅ **Approve** — no blockers (both my Minors now resolved)

The amendment is exactly the cycle_id parity fix I flagged. I diffed the PR's own files between my approved head (`9f174183`) and the new head (`270d91b4`): the only code change is the one-line cycle compare in `event_matches_query_filters`, plus a dedicated regression test.

### Minor resolved — event-side cycle_id now str-normalized
```diff
-    if cycle_id is not None and ids.get("cycle_id") != str(cycle_id):
+    if cycle_id is not None and str(ids.get("cycle_id")) != str(cycle_id):
```
This restores the exact behavior of the inline matcher this PR replaced (`str(record_cycle_id) == str(cycle_id)`): both sides are now str-normalized, so a non-string event `ids["cycle_id"]` (e.g. an int epoch) again matches a string `--cycle-id` filter. For the contractual string cycle_id it's unchanged, and the missing-cycle_id case still correctly no-matches (`str(None) == "None"` ≠ any real filter value). The shared matcher is now byte-equivalent to the inline one it replaced.

### Regression test pins the exact edge
The new `test_event_query_cycle_filter_matches_non_string_event_id` writes an event with `ids={"cycle_id": 7}` (an **int**) and a second with a string cycle_id, runs `build_event_report(..., cycle_id="7")`, and asserts the int-cycle_id event matches (`cycle.matched_events == 1`, matched event's `ids.cycle_id == 7`). That locks in the precise case that the un-normalized compare would have dropped — good, it would fail loudly if the `str()` is ever removed again.

### Tests
Ran `tests/test_live_event_query.py` + `tests/test_live_smoke_report.py` + `tests/test_live_incident_bundle.py` in a clean /tmp worktree at the amended head (`270d91b4`): **116 passed, 0 failed, 0 errors** (`event_query.py` `py_compile`-clean; +1 over the prior 115 = the new regression test). Re-ran the full matcher blast radius since the change is in the shared `event_matches_query_filters`.

Net: with this amendment, both review Minors are closed — `build_event_report` and the incident time-window report share one matcher (no drift), and that matcher is now an exact behavioral match for the inline code it replaced, including non-string cycle_ids, with a regression test. Re-approving the amended head.

PR #939, opened 2026-07-01
Summarize restart smoke plans

enarjord:
Scope:
- Add summarize_live_restart_smoke_plan() as a concise projection of the existing full live-restart-smoke-plan report.
- Add CLI --summary to emit bot count/names, phase names, planned smoke command, planned incident-bundle command, execution policy, warning/issue counts, and escalation command counts without per-bot phase detail.
- Fold PR #938 deployment/smoke evidence into the progress ledger and document planner --summary.

Behavior boundary:
- Read-only output projection only.
- Does not alter full restart plan generation, planned commands, smoke/incident-bundle behavior, execution policy, process signaling, tmux/SSH/git behavior, event producers, exchange calls, cache mutation, readiness gates, console routing, order logic, risk logic, or trading behavior.

Validation:
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_restart_smoke_plan.py -q
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/restart_smoke_plan.py src/tools/live_restart_smoke_plan.py tests/test_live_restart_smoke_plan.py
- git diff --check
- diff-only silent-handling scan for touched Python/test files: clean

PR #938, opened 2026-07-01
Plan bounded incident bundle evidence command

enarjord:
Scope:
- Add a plan-only live-incident-bundle evidence command to live-restart-smoke-plan.
- Reuse the same recent-window, event-tail, per-bot file cap, log file, log tail, and log match bounds as the planned smoke command.
- Default the planned bundle to --no-event-segments and --compact for quick failure evidence capture.
- Add --incident-bundle-output to choose the planned output path.
- Fold PR #937 deployment/smoke evidence into the progress ledger and document the new planned command.

Behavior boundary:
- Read-only planner/tooling only.
- Does not execute the incident bundle command, create bundles, signal processes, invoke tmux, SSH, git pull, start bots, contact exchanges, load credentials, add event producers, mutate caches, change readiness gates, route console output, or alter order/risk/trading behavior.

Validation:
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_restart_smoke_plan.py -q
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/restart_smoke_plan.py src/tools/live_restart_smoke_plan.py tests/test_live_restart_smoke_plan.py
- git diff --check
- diff-only silent-handling scan for touched Python/test files: clean

PR #940, opened 2026-07-01
Timestamp planned incident bundle output

enarjord:
Scope:
- Change live-restart-smoke-plan's default planned incident-bundle output from a fixed /tmp path to a UTC timestamped /tmp/passivbot_incident_bundle_restart_smoke_<utc>.tar.gz path.
- Keep --incident-bundle-output as the explicit override; empty/omitted values use the timestamped default.
- Update CLI help/tests/docs and fold PR #939 deployment/smoke evidence into the progress ledger.

Behavior boundary:
- Read-only planner output behavior only.
- Does not execute bundles, create files, alter full restart plan semantics beyond the default planned output path, signal processes, invoke tmux, SSH, git, start bots, contact exchanges, load credentials, add event producers, mutate caches, change readiness gates, route console output, order logic, risk logic, or trading behavior.

Validation:
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_restart_smoke_plan.py -q
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/live/restart_smoke_plan.py src/tools/live_restart_smoke_plan.py tests/test_live_restart_smoke_plan.py
- git diff --check
- diff-only silent-handling scan for touched Python/test files: clean

### DIFF
--- a/docs/plans/live_logging_overhaul_progress.md
+++ b/docs/plans/live_logging_overhaul_progress.md
@@ -19,7 +19,7 @@ Last updated: 2026-07-01.
 
 Current `origin/v8` logging-overhaul head:
 
-- `8ffb9a73` after PR #938, `Plan bounded incident bundle evidence command`.
+- `e3c616d4` after PR #939, `Summarize restart smoke plans`.
 
 Current review gate:
 
@@ -49,6 +49,34 @@ Retuned goal boundary:
 
 VPS5 deployment status:
 
+- Repository pulled through PR #939 at `e3c616d4`.
+- PR #939 added a read-only `live-restart-smoke-plan --summary` projection for
+  repeated operator loops. The projection keeps bot count/names, phase names,
+  planned smoke and incident-bundle commands, execution policy, signal-safety
+  strategy, escalation command counts, and warning/issue counts without dumping
+  every per-bot phase. The slice did not alter full restart plan generation,
+  planned commands, smoke/incident-bundle behavior, execution policy, process
+  signaling, tmux/SSH/git behavior, event producers, exchange calls, cache
+  mutation, readiness gates, console routing, order logic, risk logic, or
+  trading behavior.
+- PR #939 passed Hermes + Claude + Cursor + CI. Local validation covered
+  `tests/test_live_restart_smoke_plan.py`, py_compile for touched files,
+  `git diff --check`, and a diff-only silent-handling scan.
+- VPS5 pulled from `8ffb9a73` to `e3c616d4` without bot restart because the
+  deployed change was read-only planner projection tooling. The five
+  configured bots were left running.
+- A VPS5 `live-restart-smoke-plan --summary` run against
+  `/root/bots_vps5.yaml` completed with `ok=true`, five configured bots, six
+  phase names, `execute=false` for both planned commands, zero issues, and
+  bounded smoke/incident-bundle commands carrying `--event-tail-lines 2000`,
+  `--max-event-files-per-bot 2`, `--max-log-files 8`,
+  `--log-tail-lines 1200`, `--max-log-matches 20`, and `--compact`.
+- A bounded 5-minute VPS5 smoke after the pull reported `ok=true`,
+  `hard_failures=0`, clean tracked repository state at `e3c616d4`, five
+  matched expected live processes, no missing/extra/duplicate live processes,
+  no failed account-critical remote calls, `logs.max_files=8`,
+  `logs.tail_lines=1200`, and `logs.max_matches=20`. Remaining attention came
+  from known non-hard EMA readiness and HSL status/cooldown diagnostics.
 - Repository pulled through PR #938 at `8ffb9a73`.
 - PR #938 made the read-only `live-restart-smoke-plan` output include a
   bounded `live-incident-bundle` evidence command for non-clean restart/smoke
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -200,9 +200,11 @@ Monitor commands are documented in detail in [monitor.md](monitor.md). The CLI s
   credentials. The plan also includes a bounded `live-incident-bundle` command
   for failure evidence, reusing the same event/log scan limits and disabling
   event-segment copying by default; use `--incident-bundle-output PATH` to
-  choose the planned bundle path. Use planner `--summary` when you only need
-  the bot count, phase names, smoke command, and incident-bundle command without
-  every per-bot phase detail.
+  choose the planned bundle path. The default planned bundle path is a
+  timestamped `/tmp/passivbot_incident_bundle_restart_smoke_<utc>.tar.gz`
+  path to avoid overwriting prior evidence. Use planner `--summary` when you
+  only need the bot count, phase names, smoke command, and incident-bundle
+  command without every per-bot phase detail.
 - `passivbot tool live-performance-report` summarizes local live monitor event timings for
   operator performance analysis. It is read-only and does not contact exchanges. Use
   `--recent-minutes` for a time window, `--summary` for a bounded operator projection, and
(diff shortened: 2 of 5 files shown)


enarjord (commented):
**PR:** #940 — Timestamp planned incident bundle output
**Head SHA:** `06027b611a329b9b937e63867cfae4640638cc00`
**Base:** v8 (merge-base `e3c616d4` = current tip — branch is right at HEAD)
**Scope:** restart-smoke-planner library + CLI + docs — `src/live/restart_smoke_plan.py` (+20/−2), `src/tools/live_restart_smoke_plan.py` (+3/−2), `tests/test_live_restart_smoke_plan.py` (+14/−5), docs/tools.md (+5/−3), docs/plans (+30/−?). Plan-only/read-only, no live path, no producers, no verdict change.
**Verdict:** ✅ **Approve** — no blockers

Makes the planner's default `--incident-bundle-output` a timestamped path (`/tmp/passivbot_incident_bundle_restart_smoke_<UTC>.tar.gz`) instead of the fixed constant from #938, so repeated planned bundle runs don't overwrite prior evidence.

### Clean default-vs-override split
The constant is replaced by `_default_incident_bundle_output()`, which builds `/tmp/passivbot_incident_bundle_restart_smoke_<strftime %Y%m%d_%H%M%S_%f>.tar.gz`. `build_live_restart_smoke_plan`'s param defaults to `None` and resolves to the timestamped path only when it's `None`/blank (`if incident_bundle_output is None or not str(...).strip()`), so an explicit `--incident-bundle-output PATH` still wins unchanged. The CLI default also becomes `None` and the help now advertises the template `DEFAULT_INCIDENT_BUNDLE_OUTPUT_HELP` (`.../<utc>.tar.gz`), which is honest about the format without baking in a stale timestamp.

### Intentional non-determinism, correctly scoped and tested
`datetime.now(timezone.utc)` makes the auto-generated default vary per plan build — that's the point (unique output per run), and `%f` (microseconds) keeps it unique even for rapid successive builds. The non-determinism is confined to the *default output path only*; every other plan field stays deterministic, and callers who need a stable path pass one explicitly. The tests handle this correctly: a new `INCIDENT_BUNDLE_OUTPUT_PATTERN` regex (`/passivbot_incident_bundle_restart_smoke_\d{8}_\d{6}_\d{6}\.tar\.gz$`) replaces the old exact-path assertions in the default-path tests, while `..._can_override_incident_bundle_output` still asserts the exact explicit path — so both the timestamped default and the override are covered.

### Value-safe / read-only
The path stays a generic `/tmp` location with a UTC timestamp — no user-specific or sensitive information (a timestamp is not a magnitude). It's still `_display_path`'d in the plan metadata (`inputs.incident_bundle_output`, `incident_bundle.output_path`) as before, and appears raw only in the operational command string (consistent with the established planner pattern). The planner remains plan-only (`execute=False`; the CLI help reiterates "The plan never creates the bundle").

### Tests
Ran `tests/test_live_restart_smoke_plan.py` in a clean /tmp worktree at the PR head (`06027b61`): **16 passed, 0 failed, 0 errors** (both modules `py_compile`-clean). The default-path and CLI-JSON tests now regex-match the timestamped filename; the command test checks the `--output /tmp/passivbot_incident_bundle_restart_smoke_` prefix + `.tar.gz`; the explicit-override test is unchanged and green.

### Docs
docs/tools.md documents the timestamped default "to avoid overwriting prior evidence" and the `--incident-bundle-output PATH` override. docs/plans head-pointer `e3c616d4` (#939) = current tip; folds #939's deploy evidence. Faithful, value-safe, no overclaim; coin-HSL replay #777/#802 correctly still open.

Net: a small, well-scoped change that timestamps the default planned incident-bundle output to prevent overwrites — non-determinism confined to the default path and properly regex-tested, explicit override preserved and tested, value-safe (generic /tmp + UTC stamp, metadata still redacted), read-only. Approving.

enarjord:
## Review: PR #940 — Timestamp planned incident bundle output

**Scope:** full PR (head `06027b61`)

### Summary
Directly addresses the minor observation from the #938 review: the default planned incident-bundle output path is now timestamped (`/tmp/passivbot_incident_bundle_restart_smoke_<UTC YYYYMMDD_HHMMSS_ffffff>.tar.gz`, microsecond precision) rather than a fixed shared path, so repeated plan invocations no longer point at the same file by default. `--incident-bundle-output` remains the explicit override, and empty/omitted values (including an explicitly passed empty string) fall back to a freshly generated timestamped default via `.strip()` check.

### Findings
None.

### Architecture/design fit
Clean, minimal change. `_default_incident_bundle_output()` is a small pure helper computed once per `build_live_restart_smoke_plan` call. Default is moved from a module-level constant to `None` + lazy generation — correct, since a module-level constant computed at import time would have frozen the timestamp for the process lifetime and defeated the purpose. CLI default likewise changed to `None` so argparse doesn't pre-bind a stale value. Help text updated to describe the pattern (`DEFAULT_INCIDENT_BUNDLE_OUTPUT_HELP`) rather than a literal path, appropriately since the real default is now dynamic.

### Tests
- Existing assertions switched from exact-string `.endswith(...)` checks to a regex pattern (`_\d{8}_\d{6}_\d{6}\.tar\.gz$`) matching the timestamped filename, applied to both the `build_live_restart_smoke_plan` full-report test and the CLI JSON-output test.
- Ran `tests/test_live_restart_smoke_plan.py`: 16 passed, unchanged count (existing tests updated in place, no regressions).

### Verdict
**Approve.**

vigilpbclaw-hash:
PR: #940
Head SHA: 06027b611a329b9b937e63867cfae4640638cc00
Base SHA/branch: e3c616d44dff3256a69baac89ea770f8c185e3ee/v8
Scope reviewed: full PR
Verdict: approve

No Blocker/Important/Minor findings.

Intent/architecture: This PR is trying to make the read-only restart smoke planner's default planned incident-bundle output unique per invocation so repeated operator loops do not point at the same `/tmp` tarball while preserving explicit overrides. The design fits that intent: the default is generated inside `build_live_restart_smoke_plan()` only when the input is `None` or blank, avoiding an import-time timestamp that would be stale for long-lived processes, and the CLI now passes `None` rather than freezing a default. The change remains observability/planner-only and does not execute bundles, signal processes, contact exchanges, alter readiness gates, or touch trading/risk/order behavior.

Evidence reviewed:
- `src/live/restart_smoke_plan.py`: `_default_incident_bundle_output()` uses UTC timestamp + microseconds under the existing generic `/tmp/passivbot_incident_bundle_restart_smoke_*.tar.gz` pattern; explicit nonblank `incident_bundle_output` still wins and all emitted commands remain shell-quoted through the existing `_shell_join()` path.
- `src/tools/live_restart_smoke_plan.py`: argparse default changed to `None`, with help text documenting the dynamic path pattern and preserving plan-only semantics.
- `tests/test_live_restart_smoke_plan.py`: default-path assertions now regex-match the timestamped output, while the explicit override test continues to assert exact override propagation.
- Docs updates accurately describe the timestamped default and fold in PR #939 deployment/smoke evidence without expanding runtime behavior.

Validation run in `/tmp/passivbot-hermes-review-940` at head `06027b611a329b9b937e63867cfae4640638cc00`:
- `python3 -m pytest tests/test_live_restart_smoke_plan.py -q` — 16 passed, 1 pre-existing config warning about `asyncio_mode`.
- `python3 -m py_compile src/live/restart_smoke_plan.py src/tools/live_restart_smoke_plan.py tests/test_live_restart_smoke_plan.py` — passed.
- Merge check: `git merge --no-commit --no-ff 06027b611a329b9b937e63867cfae4640638cc00` from current `origin/v8` (`e3c616d44dff3256a69baac89ea770f8c185e3ee`) — automatic merge succeeded.

PR #1049, opened 2026-07-04
Validate optimizer overrides

enarjord:
## Summary\n- reject unknown optimize.enable_overrides names before optimize startup and from direct override calls\n- restore trailing_grid_v7 forward_tp_grid / backward_tp_grid behavior by ordering close grid markup start/end\n- keep TP-grid overrides v7-only; trailing_martingale users should express forward/backward grids through threshold_we_weight bounds\n\n## Validation\n- ./venv/bin/python -m pytest tests/optimization/test_optimize.py::TestIndividualToConfig tests/optimization/test_optimizer_warmup.py\n- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Optimizer overrides now reject unknown `optimize.enable_overrides` names before
+  the run starts, and `forward_tp_grid` / `backward_tp_grid` now reorder
+  `trailing_grid_v7` close-grid markup bounds as intended.
 - Optimizer stepped bounds now stay on the configured grid for fractional steps
   such as `0.25`, `0.125`, and `0.0025`, avoiding off-grid candidate values in
   DEAP, pymoo repair, seed conversion, and result hashing.
--- a/src/optimize.py
+++ b/src/optimize.py
@@ -153,7 +153,7 @@ except ImportError:  # pragma: no cover - allow import in minimal test envs
     tools = algorithms = None
 import math
 import fcntl
-from optimizer_overrides import optimizer_overrides
+from optimizer_overrides import optimizer_overrides, validate_optimizer_overrides
 from opt_utils import make_json_serializable, generate_incremental_diff, round_floats, quantize_floats
 from limit_utils import expand_limit_checks, compute_limit_violation
 from pareto_store import ParetoStore
@@ -2851,6 +2851,7 @@ async def main():
         verbose=False,
         raw_snapshot=raw_snapshot,
     )
+    validate_optimizer_overrides(config.get("optimize", {}).get("enable_overrides", []))
     config_logging_value = get_optional_config_value(config, "logging.level", None)
     effective_log_level = resolve_log_level(args.log_level, config_logging_value, fallback=1)
     if effective_log_level != initial_log_level:
--- a/src/optimizer_overrides.py
+++ b/src/optimizer_overrides.py
@@ -4,6 +4,17 @@ from config.access import require_config_value
 from config.shared_bot import BOT_SHARED_GROUPS
 from config.strategy import DEFAULT_STRATEGY_KIND, normalize_strategy_kind
 
+TRAILING_GRID_V7_STRATEGY_KIND = "trailing_grid_v7"
+
+KNOWN_OPTIMIZER_OVERRIDES = frozenset(
+    {
+        "backward_tp_grid",
+        "forward_tp_grid",
+        "lossless_close_trailing",
+        "mirror_short_from_long",
+    }
+)
+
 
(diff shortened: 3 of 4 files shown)


enarjord (commented):
**PR:** #1049 — Validate optimizer overrides
**Head SHA:** `98298dcf395762b35fda87bdcb900a4c6402de6e`
**Base:** v8 (merge-base `3f820248` = current tip — branch is right at HEAD)
**Scope:** `src/optimizer_overrides.py` (+74/−13), `src/optimize.py` (+3/−1: startup validation), tests (+40), CHANGELOG. Offline optimizer only.
**Verdict:** ✅ **Approve** — no blocking findings; 1 informational note below.

Two fixes, both closing silent-misbehavior holes:

1. **Unknown override names now fail loudly, twice.** `validate_optimizer_overrides` rejects unknown `optimize.enable_overrides` values against a frozen allowlist at startup (`main()`, right after config load) and again inside `optimizer_overrides`. The old fallback was worse than it looked: `print(f"Unknown override: ...")` followed by `return config` **inside the loop** — an unknown name didn't just no-op, it silently skipped every override listed after it. The dead `"example"` branch is gone too.

2. **`forward_tp_grid` / `backward_tp_grid` get real semantics.** Previously both branches only called `_require_trailing_martingale_side` — a validation-only no-op that additionally asserted the *wrong* strategy kind for a close-grid-markup concept. Now they require `trailing_grid_v7` and deterministically reorder `close.grid_markup_start/end` (forward = ascending, backward = descending, via min/max so the pair is normalized regardless of input order; equal values degenerate harmlessly).

Checked for regressions along the way:
- `_record_individual_result` calls `optimizer_overrides(overrides_list, entry, None)` — with a tp-grid override enabled and `pside=None`, the path lookup fails just as the old `_require_trailing_martingale_side` would have; no new failure mode introduced there.
- The per-candidate validation cost is a set comprehension over a handful of names — negligible in the hot path.

### Informational: configs carrying these overrides under `trailing_martingale` now abort
Any existing config that lists `forward_tp_grid`/`backward_tp_grid` with `live.strategy_kind = trailing_martingale` previously "worked" (validated and did nothing); it now raises `optimizer override requires live.strategy_kind = 'trailing_grid_v7'`. That's the correct fail-loud outcome for an override that never did anything, but users with stale override lists will need to remove them — worth keeping in mind if support questions show up post-merge. The CHANGELOG covers the intent.

### Tests (clean /tmp worktree at `98298dcf`)
`tests/optimization/test_optimize.py` + `test_backends.py`: **195 passed, 0 failed**. New tests pin unknown-name rejection, both reorderings against a `trailing_grid_v7` config, and strategy-kind rejection for other strategies.

Net: the override mechanism now either does exactly what its names promise or refuses to run — no silent skips, no placeholder branches. Approving.

vigilpbclaw-hash:
PR: #1049
Head SHA: 98298dcf395762b35fda87bdcb900a4c6402de6e
Base SHA/branch: 3f8202485cfdb0fc1c37e3c12edf8d8c81b9387c/v8
Scope reviewed: full PR
Verdict: approve

Supersedes: no previous Hermes Continuous Reviewer review was found for this PR.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this PR makes optimizer overrides fail-loud and gives the TP-grid override names real trailing-grid-v7 semantics. That is a good fit for optimizer safety: unknown override names and previously no-op grid-direction names can silently change optimization behavior, so validating once at startup and again at direct override application is preferable to best-effort continuation. Keeping `forward_tp_grid`/`backward_tp_grid` scoped to `trailing_grid_v7` also avoids pretending those close-grid concepts apply to `trailing_martingale`.

Evidence reviewed:
- `src/optimizer_overrides.py`: fixed allowlist validation, direct-call validation, removal of the dead `example`/unknown-print path, and deterministic close `grid_markup_start/end` ordering for `forward_tp_grid` and `backward_tp_grid`.
- `src/optimize.py`: startup validation of `optimize.enable_overrides` immediately after config load.
- `tests/optimization/test_optimize.py`: coverage for unknown names, both TP-grid orderings, and rejection under the wrong strategy kind.

Validation run:
- `git diff --check origin/v8...98298dcf395762b35fda87bdcb900a4c6402de6e` — passed.
- In `/tmp/passivbot-hermes-continuous` with a uv Python 3.12 venv, dependencies installed, and `passivbot-rust` rebuilt/installed: `python -m pytest tests/optimization/test_optimize.py::TestIndividualToConfig::test_optimizer_overrides_reject_unknown_names tests/optimization/test_optimize.py::TestIndividualToConfig::test_trailing_grid_v7_tp_grid_overrides_reorder_close_markup tests/optimization/test_optimize.py::TestIndividualToConfig::test_trailing_grid_v7_tp_grid_overrides_reject_other_strategies -q` — passed (4 passed).

PR #1056, opened 2026-07-04
Preserve all_results diff deletions

enarjord:
## Summary
- add an explicit deletion marker to compressed optimizer result diffs
- teach all_results replay to remove top-level and nested deleted keys
- share deletion-aware diff logic between generate_diffs and generate_incremental_diff

## Tests
- ./venv/bin/python -m pytest tests/test_opt_utils.py tests/optimization/test_pymoo_backend.py::test_run_backend_writes_readable_result_artifacts tests/optimization/test_optimize.py::test_result_recorder_all_results_write_failure_is_fatal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Compressed `all_results.bin` optimizer history now preserves deleted keys
+  during replay, preventing stale fields such as prior candidate errors from
+  leaking into later entries.
 - Suite scenarios now reject unknown scenario fields before running, catching
   typos such as `coin` instead of silently ignoring them.
 - Optimizer overrides now reject unknown `optimize.enable_overrides` names before
--- a/src/opt_utils.py
+++ b/src/opt_utils.py
@@ -5,6 +5,29 @@ import msgpack
 from typing import Any
 import passivbot_rust as pbr
 
+_DIFF_DELETE_KEY = "__passivbot_diff_delete__"
+_DIFF_DELETE_MARKER = {_DIFF_DELETE_KEY: True}
+
+
+def _is_diff_delete_marker(value: Any) -> bool:
+    return isinstance(value, dict) and value == _DIFF_DELETE_MARKER
+
+
+def _dict_diff(d1: dict, d2: dict) -> dict:
+    diff = {}
+    for k in d1.keys() - d2.keys():
+        diff[k] = _DIFF_DELETE_MARKER.copy()
+    for k in d2:
+        if k not in d1:
+            diff[k] = d2[k]
+        elif isinstance(d2[k], dict) and isinstance(d1.get(k), dict):
+            nested = _dict_diff(d1[k], d2[k])
+            if nested:
+                diff[k] = nested
+        elif d1[k] != d2[k]:
+            diff[k] = d2[k]
+    return diff
+
 
 def dominates(p0, p1):
     better_in_one = False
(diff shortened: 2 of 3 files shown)


enarjord (commented):
**PR:** #1056 — Preserve all_results diff deletions
**Head SHA:** `39b356285ba5d9bdd51679d18806798fcb3d88b9`
**Base:** v8 (merge-base `b6a9a606`)
**Scope:** `src/opt_utils.py` (+57/−30: shared `_dict_diff` with deletion markers), `tests/test_opt_utils.py` (+55), CHANGELOG. Offline optimizer results serialization only.
**Verdict:** ✅ **Approve** — no findings.

Real correctness bug in the compressed `all_results.bin` round-trip: both (previously duplicated) `dict_diff` implementations only recorded added/changed keys — key **deletions** were silently dropped, so replaying diffs via `deep_updated`/`apply_diffs` carried stale fields forward indefinitely. The CHANGELOG's example is the nasty concrete case: an invalid candidate's `metrics.error` field would bleed into every subsequent replayed entry until some entry happened to overwrite it. The fix records `{"__passivbot_diff_delete__": true}` for keys in `d1.keys() - d2.keys()` (recursively, so nested deletions work), and `deep_updated` drops marker-valued keys on replay.

Checked beyond the diff itself:
- **Other `deep_updated` consumers are unaffected:** `optimize.py` uses it for seed/bounds config merging (non-diff dicts) — the marker-skip branch only fires on a value exactly equal to the sentinel dict, which cannot occur in bot/live/bounds configs. Legacy `all_results.bin` files (no markers) replay identically to before.
- **Deduplication bonus:** the two identical inline `dict_diff` closures in `generate_diffs` and `generate_incremental_diff` are consolidated into one module-level `_dict_diff`, so writer and replay semantics can't drift again.
- One-way format note (expected for a bug fix): files written by this version contain markers that older readers would surface as literal dicts — new reader handles both, old files are fine.

### Tests (clean /tmp worktree at `39b35628`)
`tests/test_opt_utils.py` + `tests/optimization/test_optimize.py`: **193 passed, 0 failed**. New tests pin top-level and nested deletion replay for both diff generators, plus an end-to-end `load_results` round-trip through a real msgpack file.

### Docs
CHANGELOG entry accurate.

Net: compressed results history now replays exactly what was recorded, deletions included, with the duplicated diff logic consolidated. Approving.

vigilpbclaw-hash:
PR: #1056
Head SHA: 39b356285ba5d9bdd51679d18806798fcb3d88b9
Base SHA/branch: b6a9a6060355e1b022daa4abadbd47c8b2cca23a/v8
Scope reviewed: full PR
Verdict: approve

- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this PR fixes compressed optimizer `all_results.bin` replay so keys that disappear between candidates are represented explicitly instead of being silently retained from prior records. Centralizing the old duplicated diff implementations into `_dict_diff()` is the right design boundary: both full-list and incremental writers now share deletion semantics, while `deep_updated()` remains the single replay path. The sentinel marker is narrow and exact-match based, so normal config/result dictionaries continue to merge as before unless they exactly carry the internal deletion marker.

Evidence reviewed:
- `src/opt_utils.py`: `_dict_diff()` now records top-level and nested deletions, `generate_diffs()` and `generate_incremental_diff()` share it, and `deep_updated()` skips deletion markers during replay.
- `tests/test_opt_utils.py`: regressions cover top-level deletion, nested deletion, `load_results()` round-trip through msgpack, and `generate_diffs()` replay.
- CHANGELOG entry matches the user-facing optimizer history fix.

Validation run in `/tmp/pb-review-1056`:
- `git diff --check b6a9a6060355e1b022daa4abadbd47c8b2cca23a...HEAD` — passed.
- Built `passivbot_rust` into a temporary Python 3.12 review venv with `maturin develop --release` so optimizer tests could import native bindings.
- `PYTHONPATH=src python -m pytest -q tests/test_opt_utils.py tests/optimization/test_pymoo_backend.py::test_run_backend_writes_readable_result_artifacts tests/optimization/test_optimize.py::test_result_recorder_all_results_write_failure_is_fatal` — passed (`6 passed`).

PR #1048, opened 2026-07-04
Add optional optimizer seed control

enarjord:
## Summary
- add optimize.seed with default null, normalized from null/auto/random or integer values
- keep default optimizer RNG behavior randomized while allowing fixed integer seeds for diagnostics/determinism
- seed parent RNGs before backend setup, seed worker RNGs independently, and pass the configured seed to pymoo instead of hardcoded seed=1

## Tests
- ./venv/bin/python -m pytest tests/optimization/test_backends.py tests/optimization/test_pymoo_backend.py
- ./venv/bin/python -m pytest tests/test_config_pipeline.py
- ./venv/bin/python -m pytest tests/test_config_utils_helpers.py
- ./venv/bin/python -m pytest tests/optimization/test_backends.py tests/optimization/test_pymoo_backend.py tests/test_config_pipeline.py tests/test_config_utils_helpers.py

PR #1071, opened 2026-07-04
Install DEAP evaluator once per worker

enarjord:
## Summary
- install the DEAP evaluator and overrides into worker globals from the pool initializer
- register a module-level DEAP evaluation function so per-task payloads no longer carry a bound evaluator method
- add focused tests for the initializer/global evaluation path and task-function picklability

## Tests
- ./venv/bin/python -m pytest tests/optimization/test_backends.py
- ./venv/bin/python -m py_compile src/optimization/backends/deap_backend.py tests/optimization/test_backends.py
- git diff --check

### DIFF
--- a/src/optimization/backends/deap_backend.py
+++ b/src/optimization/backends/deap_backend.py
@@ -7,7 +7,7 @@ import os
 import pickle
 import random
 from copy import deepcopy
-from typing import Any, Callable
+from typing import Any, Callable, Sequence
 
 import numpy as np
 
@@ -33,6 +33,8 @@ from optimization.random_seed import seed_worker_rngs
 from config.scoring import engine_space_fitness_weights
 
 DEFAULT_DEAP_POPULATION_SIZE = 500
+_DEAP_WORKER_EVALUATOR = None
+_DEAP_WORKER_OVERRIDES_LIST: list[Any] = []
 
 
 def _resolve_deap_population_size(config: dict[str, Any]) -> int:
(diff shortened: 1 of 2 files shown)


vigilpbclaw-hash:
PR: #1071
Head SHA: 88379c22e479b29d056035536a3aedd920f5a6d6
Base SHA/branch: aeff33ed694dc373d03fe5c4681582335b4638cc/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: the PR is trying to reduce DEAP multiprocessing task-payload size by installing the evaluator and overrides once per worker, then submitting a module-level task function. This fits Python multiprocessing better than repeatedly pickling a bound evaluator method per individual, and it stays within the existing DEAP backend abstraction: evaluation still flows through the same evaluator object and overrides list, while seed/signal worker initialization remains centralized.

Evidence reviewed:
- `src/optimization/backends/deap_backend.py`: worker globals are set from the pool initializer; `_evaluate_deap_worker` fails loudly if used without initializer state; `toolbox.evaluate` is now a module-level function passed to `pool.map`/`apply_async`.
- `tests/optimization/test_backends.py`: covers initializer-installed globals and picklability of the module-level evaluator task function.

Validation run:
- `python3 -m py_compile src/optimization/backends/deap_backend.py tests/optimization/test_backends.py` — passed.
- `python3 -m pytest tests/optimization/test_backends.py -q` — could not run in this cron environment because dependencies are missing (`ModuleNotFoundError: No module named 'ccxt'`) during test collection.

enarjord (commented):
**PR:** #1071 — Install DEAP evaluator once per worker
**Head SHA:** `88379c22e479b29d056035536a3aedd920f5a6d6`
**Base:** v8 (merge-base `782b1679` = current tip)
**Scope:** `src/optimization/backends/deap_backend.py` (+30/−3), `tests/optimization/test_backends.py` (+15). Offline optimizer only.
**Verdict:** ✅ **Approve** — no blocking findings; local test-run caveat below.

Sound IPC optimization mirroring the established pymoo pattern: previously `toolbox.register("evaluate", evaluator_for_pool.evaluate, overrides_list=...)` bound the evaluator method directly, so **every** `pool.apply_async(toolbox.evaluate, (ind,))` task message pickled the bound method — dragging the whole evaluator (config, bounds, key paths) across the pipe per candidate. Now the evaluator + overrides ship **once per worker** via the pool initializer (`_initialize_deap_worker` → `_set_deap_worker_globals`, exactly the `_PYMOO_WORKER_*` globals pattern from `problem.py`), and tasks carry only the individual, with `_evaluate_deap_worker` failing loudly if a worker was somehow not initialized.

Verified the safety-critical invariant: both `toolbox.evaluate` call sites at this head (`deap_backend.py:226` and `optimize.py:980` in `ea_mu_plus_lambda_stream`) go through `pool.apply_async`, i.e. execute **in workers** where the globals are installed — there is no parent-side invocation to hit the RuntimeError. Shared duplicate-guard state is unaffected (`seen_hashes`/counters are Manager proxies inside the evaluator, shipped once and still shared). The initializer keeps its seed-then-sigint ordering from #1048.

**Caveat:** I could not execute the suites locally this tick — the review environment's prebuilt rust extension now fails the source-fingerprint gate after recent `passivbot-rust/` merges (documented standing env limit; no rebuilds performed in the review clone). The analysis above is from reading all call sites; the PR's new worker-installation test should be relied on via the author/CI run.

Net: per-candidate evaluator pickling replaced by once-per-worker installation, symmetric with pymoo. Approving.

PR #1072, opened 2026-07-04
Avoid forced optimizer data pre-copy

enarjord:
## Summary
- avoid unconditional pre-copying of optimizer HLCV and BTC arrays before shared-memory registration
- keep dtype/C-order normalization for unsuitable inputs via np.asarray
- add a regression test proving already-contiguous float64 arrays are handed to the shared-array manager by identity

## Tests
- ./venv/bin/python -m pytest tests/optimization/test_optimize.py::TestValidateArray
- ./venv/bin/python -m py_compile src/optimize.py tests/optimization/test_optimize.py
- git diff --check

### DIFF
--- a/src/optimize.py
+++ b/src/optimize.py
@@ -349,10 +349,10 @@ def _register_exchange_data(
     config["backtest"]["coins"][exchange] = coins
     msss[exchange] = mss
     validate_array(hlcvs, "hlcvs")
-    hlcvs_array = np.array(hlcvs, dtype=np.float64, copy=True, order="C")
+    hlcvs_array = np.asarray(hlcvs, dtype=np.float64, order="C")
     hlcvs_spec, _ = array_manager.create_from(hlcvs_array)
     hlcvs_specs[exchange] = hlcvs_spec
-    btc_usd_array = np.array(btc_usd_prices, dtype=np.float64, copy=True, order="C")
+    btc_usd_array = np.asarray(btc_usd_prices, dtype=np.float64, order="C")
     validate_array(btc_usd_array, f"btc_usd_data for {exchange}", allow_nan=False)
     btc_usd_spec, _ = array_manager.create_from(btc_usd_array)
     btc_usd_specs[exchange] = btc_usd_spec
--- a/tests/optimization/test_optimize.py
+++ b/tests/optimization/test_optimize.py
@@ -31,6 +31,7 @@ from optimize import (
     _normalize_optional_bool_flag,
     _optimizer_exit_code,
     _record_individual_result,
+    _register_exchange_data,
     _resolve_cli_limits_override,
     _set_candidate_metrics,
     _terminate_optimizer_pool,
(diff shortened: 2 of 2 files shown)


vigilpbclaw-hash:
PR: #1072
Head SHA: 08db31cdcf26d068fe1aee217ac3081f7bce1047
Base SHA/branch: 782b16796cfe9c13ba5ea3bcd44d1a42563ca91e/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this PR avoids an unconditional optimizer-side NumPy pre-copy before data is handed to `SharedArrayManager`, while preserving dtype/C-order normalization for unsuitable inputs. That is a good fit for the existing shared-array abstraction: `SharedArrayManager.create_from()` remains the single place that allocates/copies into shared memory, and `_register_exchange_data()` no longer adds an avoidable extra parent-process copy for already-suitable arrays.

Evidence reviewed:
- `src/optimize.py::_register_exchange_data`: switches `np.array(..., copy=True, order="C")` to `np.asarray(..., dtype=np.float64, order="C")` for HLCV and BTC arrays; validation still rejects invalid NaN/inf data before shared-memory registration.
- `src/shared_arrays.py::SharedArrayManager.create_from`: still makes/uses a contiguous view and copies into the shared memory segment, so the change only removes the redundant pre-copy.
- `tests/optimization/test_optimize.py`: adds a regression test proving already-contiguous float64 arrays are passed to the manager by identity.

Validation run:
- `python3 -m py_compile src/optimize.py tests/optimization/test_optimize.py` — passed.
- `python3 -m pytest tests/optimization/test_optimize.py::TestValidateArray -q` — could not run in this cron environment because dependencies are missing (`ModuleNotFoundError: No module named 'pandas'`) during test collection.

enarjord (commented):
**PR:** #1072 — Avoid forced optimizer data pre-copy
**Head SHA:** `08db31cdcf26d068fe1aee217ac3081f7bce1047`
**Base:** v8 (merge-base `782b1679` = current tip)
**Scope:** `src/optimize.py` (+4/−2 in `_register_exchange_data`), `tests/optimization/test_optimize.py` (+32). Offline optimizer memory behavior only.
**Verdict:** ✅ **Approve** — no blocking findings; local test-run caveat below.

Correct removal of a redundant full copy on the biggest arrays in the process: `_register_exchange_data` built `np.array(hlcvs, dtype=float64, copy=True, order="C")` immediately before handing it to `SharedArrayManager.create_from` — and I verified `create_from` **always copies** (`np.ascontiguousarray` + fresh `SharedMemory` + `np.copyto`), so the explicit pre-copy just doubled peak transient memory for the multi-GB HLCV matrix during registration. `np.asarray(..., dtype, order="C")` is the right replacement: it's a no-op view when the source is already float64/C-contiguous (the common case) and converts only when needed; the shared-memory segment is a fresh copy either way, so no aliasing risk downstream — consumers only ever see the shared view, and the source array is not mutated afterwards in this function. `validate_array` ordering is preserved (hlcvs validated before conversion, btc after — as before).

**Caveat:** suites could not be executed locally this tick — the review environment's prebuilt rust extension now fails the source-fingerprint gate after recent `passivbot-rust/` merges (documented standing env limit). Verification above is from reading `shared_arrays.create_from` and the registration path; the PR's new tests pin the no-extra-copy behavior on the author/CI side.

Net: registration peak memory drops by one full dataset copy with byte-identical shared-memory contents. Approving.

PR #1103, opened 2026-07-05
Remove shadowed legacy HSL methods

enarjord:
## Summary
- remove legacy Passivbot HSL methods that are shadowed by the active passivbot_hsl bindings later in the class
- keep the shared PnL coverage helpers in passivbot.py
- add an AST regression test proving HSL behavior is bound from passivbot_hsl rather than duplicated in Passivbot method defs

## Validation
- python3 -m py_compile src/passivbot.py tests/test_passivbot_hsl_bindings.py
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/passivbot.py tests/test_passivbot_hsl_bindings.py
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_passivbot_hsl_bindings.py -q
- git diff --check

## Baseline failures checked
These fail the same way on clean origin/v8, so they are not caused by this cleanup:
- tests/test_hsl_coin_mode.py::test_coin_hsl_check_tp_only_orange_skips_flat_symbols
- tests/test_unstucking_safeguards.py::test_hard_stop_initialize_from_history_terminal_stop_sets_latch
- tests/test_unstucking_safeguards.py::test_live_orchestrator_passes_merged_entry_cooldown_delta_anchor


vigilpbclaw-hash:
PR: #1103
Head SHA: 15f4e628ac79753a05cb43bcdfb32200b6eff8c5
Base SHA/branch: c0e6068ed14e8a8bad7a515ea2975d29a615cbd2/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: `tests/test_passivbot_hsl_bindings.py:13` is a useful AST guard for the primary active HSL bindings, but it is intentionally narrow: it checks that selected assignment names exist and that no broad `_equity_hard_stop*` method definitions remain in `Passivbot`. It does not runtime-import `Passivbot` or exhaustively prove every removed legacy helper maps to `passivbot_hsl`. Non-blocking because the removed definitions were shadowed by the existing later class-body assignments, and a static audit found the removed active helpers are still assigned from `pb_hsl`; the only removed helper without a current binding was `_equity_hard_stop_wait_for_cooldown`, which has no remaining references.

Intent/architecture pass: this PR is a cleanup to remove obsolete in-class HSL implementations from `src/passivbot.py` and leave `passivbot_hsl` as the single source of HSL behavior. That is the right direction architecturally: HSL/risk behavior is sensitive, and keeping duplicate shadowed implementations in the large `Passivbot` class creates future maintenance debt and review ambiguity. The active runtime contract remains the existing `pb_hsl.*` class-body bindings, so the change is behavior-neutral rather than a risk-logic rewrite.

Evidence reviewed:
- `src/passivbot.py`: removed 34 legacy HSL-related method definitions before the existing `pb_hsl` binding block; verified the active binding block still provides the HSL entry points used by startup/check/supervisor/orange-overlay paths.
- `tests/test_passivbot_hsl_bindings.py`: new AST regression ensuring `Passivbot` no longer defines duplicate HSL methods and still contains the key `passivbot_hsl` assignment names.
- Static parity script over the diff: extracted removed `def` names, compared them against current class assignment targets and `src/passivbot_hsl.py` definitions. All removed active helpers are still assigned from `pb_hsl`; `_equity_hard_stop_wait_for_cooldown` is not assigned in current code but also has no remaining references.
- Merge check: PR head merges cleanly into current `origin/v8` at `c0e6068ed14e8a8bad7a515ea2975d29a615cbd2`.

Validation run:
- `git diff --check c0e6068ed14e8a8bad7a515ea2975d29a615cbd2...HEAD` — passed.
- `python3 -m py_compile src/passivbot.py tests/test_passivbot_hsl_bindings.py` — passed.
- `python3 -m pytest tests/test_passivbot_hsl_bindings.py -q` — passed (1 passed; pytest warned about an unknown `asyncio_mode` config option in this environment).
- Attempted direct `import passivbot` binding identity smoke check, but this review environment does not have runtime dependency `ccxt` installed, so that import-only check could not run here.

enarjord (commented):
**PR:** #1103 — Remove shadowed legacy HSL methods
**Head SHA:** `15f4e628ac79753a05cb43bcdfb32200b6eff8c5`
**Base:** v8 (merge-base `c0e6068e` = current tip — branch is right at HEAD)
**Scope:** `src/passivbot.py` (−953: remove 34 shadowed in-class HSL method defs), `tests/test_passivbot_hsl_bindings.py` (+43, new ast guard). MAX-care area (HSL, contract c) but behavior-preserving dead-code removal.
**Verdict:** ✅ **Approve** — no blocking findings; verified behavior-identical.

This removes 34 legacy HSL method definitions from the `Passivbot` class body that were already dead: the class rebinds the active implementations from `passivbot_hsl` (`import passivbot_hsl as pb_hsl`) in a block of class-level assignments (`_equity_hard_stop_X = pb_hsl._equity_hard_stop_X`, ...). Because those assignments come **after** the in-class `def`s in the class body, Python's later-binding-wins rule means the `pb_hsl` bindings were always the effective methods and the in-class defs never executed. Removing them changes nothing at runtime and eliminates a real drift hazard (two copies of HSL logic where edits to the shadowed copy would silently no-op).

**Verified the removal is safe (behavior-preserving):**
- **All removed defs sit above the binding block.** In the base, the binding block (`_hsl_psides = pb_hsl._hsl_psides ...`) starts at L2671; in the new head it's at L1718 — a shift of exactly 953, the full removed line count. That arithmetic proves every removed line is above the binding block, so each removed def was shadowed by a later `pb_hsl` binding, never active.
- **33 of 34 removed methods have a `pb_hsl` binding** (checked every removed name against `pb_hsl.<name>` references, including multi-line and `staticmethod(...)`-wrapped bindings) — those are shadowed duplicates of the active `passivbot_hsl` functions.
- **The 1 exception, `_equity_hard_stop_wait_for_cooldown`, is fully dead** — not bound from `pb_hsl`, not defined in `passivbot_hsl.py`, and **zero references** anywhere in `src/`/`tests/` (its only reference at the base was its own now-removed def). Safe to drop; it was an orphaned method.
- No active code path references any removed name post-removal (grepped `wait_for_cooldown` and the `_equity_hard_stop*` set across the tree).

**New test is a good anti-regression guard.** `test_passivbot_uses_passivbot_hsl_module_for_hsl_methods` `ast`-parses `src/passivbot.py` and asserts (a) there are **no** in-class `def`s for `_equity_hard_stop*` / `_calc_upnl_sum_strict` / `_apply_equity_hard_stop_orange_overlay` — so a future accidental re-introduction of a shadowed duplicate fails CI — and (b) the key HSL methods (`_equity_hard_stop_check`, `_equity_hard_stop_check_coin`, `_equity_hard_stop_run_red_supervisor`, `_equity_hard_stop_run_coin_red_supervisor`, `_apply_equity_hard_stop_orange_overlay`) are present as class-level bindings. Pure-Python (ast only) → I ran it locally: **1 passed, 0 failed**.

### Minor (optional): no CHANGELOG
Purely internal refactor with no user-facing behavior change, so a CHANGELOG entry isn't required; a one-line "internal: removed dead shadowed HSL method definitions" note would still be nice for maintainers tracing history, but non-blocking.

Net: a safe, behavior-identical cleanup that deletes 953 lines of shadowed-dead HSL duplicates (removing a genuine silent-drift footgun) and adds a structural test to keep them from creeping back. Approving `15f4e628`.

PR #1104, opened 2026-07-05
Reset coin HSL episode after nonpanic flatten

enarjord:
Summary:
- reset live coin-HSL current-episode realized-PnL and runtime drawdown when fill replay proves an ordinary non-panic flatten before a later re-entry
- keep panic-marker cooldown/no-restart handling unchanged
- update the fable-audit action plan with this partial coin-HSL episode-anchoring slice

Validation:
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/passivbot_hsl.py tests/test_hsl_coin_mode.py
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_hsl_coin_mode.py::test_coin_hsl_history_replay_resets_current_episode_after_nonpanic_flatten -q
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_hsl_coin_mode.py -q -k 'not test_coin_hsl_check_tp_only_orange_skips_flat_symbols'
- git diff --check

Note: full tests/test_hsl_coin_mode.py still has the known origin/v8 baseline failure test_coin_hsl_check_tp_only_orange_skips_flat_symbols, reproduced before this PR.

### DIFF
--- a/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
+++ b/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
@@ -820,6 +820,11 @@ Remaining implementation details:
       cooldown rules.
       Current panic eligibility should be based on the current trading episode;
       terminal no-restart accounting remains broader.
+      Partial: live coin-HSL replay now resets current-episode realized-PnL
+      and runtime drawdown state when fill replay proves a coin+pside flattened
+      by an ordinary, non-panic close before a later re-entry. Panic-marker
+      cooldown/no-restart handling is unchanged; broader terminal no-restart
+      accounting from canonical RED timestamps remains future work.
 - [ ] HSL replay performance/readiness slice.
       Persist verified non-authoritative HSL time series/checkpoints, add doctor
       tools, prioritize held scopes, keep timing/source evidence, and move dense
@@ -830,8 +835,11 @@ Remaining implementation details:
       orders.
       Python should reconcile ideal vs actual orders, not re-decide trading
       policy where Rust can own it.
-- [ ] Dead legacy HSL cleanup after parity tests cover active
+- [x] Dead legacy HSL cleanup after parity tests cover active
       `src/passivbot_hsl.py`.
+      Implemented: removed shadowed legacy HSL method definitions from
+      `src/passivbot.py`; active behavior is bound from `src/passivbot_hsl.py`
+      with an AST regression guard.
 
 ## Non-Goals For First PRs
 
--- a/src/passivbot_hsl.py
+++ b/src/passivbot_hsl.py
@@ -2893,10 +2893,12 @@ async def _equity_hard_stop_initialize_coin_from_history(self) -> None:
 
                 replay_event_idx = 0
                 replay_size = 0.0
+                replay_was_nonflat = False
+                replay_flattened_at_ms: Optional[int] = None
                 ignored_panic_marker_timestamps: set[int] = set()
 
                 def replay_size_at(row_ts_ms: int) -> float:
-                    nonlocal replay_event_idx, replay_size
+                    nonlocal replay_event_idx, replay_size, replay_flattened_at_ms
                     boundary_ts_ms = int(row_ts_ms) + 60_000
                     while replay_event_idx < len(replay_events):
                         event_ts, action, qty = replay_events[replay_event_idx]
@@ -2906,6 +2908,8 @@ async def _equity_hard_stop_initialize_coin_from_history(self) -> None:
                             replay_size += qty
                         else:
                             replay_size = max(0.0, replay_size - qty)
+                            if replay_size <= 1e-12:
+                                replay_flattened_at_ms = int(event_ts)
                         replay_event_idx += 1
                     return float(replay_size)
 
@@ -2948,6 +2952,7 @@ async def _equity_hard_stop_initialize_coin_from_history(self) -> None:
                     )
                     if not require_coin_timeline_value:
                         continue
+                    replay_position_size = replay_size_at(ts)
                     abs_realized = _equity_hard_stop_history_coin_value(
                         row,
                         "realized_pnl_by_coin_pside",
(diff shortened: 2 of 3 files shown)


enarjord (commented):
**PR:** #1104 — Reset coin HSL episode after nonpanic flatten
**Head SHA:** `5917b31a0fb6a836a22c18f6ebdf2940038d681a`
**Base:** v8 (merge-base `3c597bd9` = current tip — branch is right at HEAD)
**Scope:** `src/passivbot_hsl.py` (+35, coin-HSL history replay episode reset), `tests/test_hsl_coin_mode.py` (+90), audit doc. MAX-care: coin-HSL replay, directly in the #777/#802 9-Goal contract (c).
**Verdict:** ✅ **Approve** — no blocking findings; correctly gated and matches the protective-stop safety model.

This implements the audit item "current panic eligibility should be based on the current trading episode" for the coin-HSL replay path: when fill replay **proves** a coin+pside flattened via an ordinary, non-panic close before a later re-entry, the current episode's realized-PnL/drawdown state is reset so a resolved historical RED doesn't make the fresh episode panic-eligible ("historical-RED-no-panic-if-now-green"). Verified against the 9-Goal contract:

**Correct, conservative gating (fail-safe on missing proof).** The reset fires only when all hold: `not replay_ambiguous` (the flatten is proven, not inferred), `replay_was_nonflat and not replay_is_nonflat` (went non-flat → flat), `replay_flattened_at_ms < ts + 60_000` (flatten within the row window), and `marker is None` (a non-panic row). When any is uncertain it does **not** reset — which is the safe direction for a protective stop: failing to reset preserves a real RED (over-protective, safe), whereas wrongly resetting would suppress a legitimate RED (under-protective, dangerous). The two sibling tests confirm the ambiguous path fails loud (`pytest.raises(ValueError, match="unrealized_pnl_by_coin_pside")`) rather than silently resetting.

**Panic accounting untouched.** The reset is inside the `if marker is None:` branch; rows carrying a panic marker fall through to the unchanged panic-handling path. So panic-marker cooldown / terminal no-restart accounting is unaffected — the audit note correctly says "Panic-marker cooldown/no-restart handling is unchanged; broader terminal no-restart accounting from canonical RED timestamps remains future work."

**Reuses established reset machinery.** `_equity_hard_stop_reset_coin_after_restart` (existing method, base L3465) and `reset_rolling_window()` (the local closure already used at three other replay sites) are reused — not an ad-hoc reset — then state is re-fetched (`state = self._hsl_coin_state(...)`) and `pnl_reset_timestamp_ms` is set to `flatten_ts + 1` so subsequent replay computes drawdown from the fresh episode boundary. The `logging.info` is value-bearing only in the sense of a timestamp (not drawdown/equity/pnl magnitudes) and is off-console.

**The moved `replay_position_size = replay_size_at(ts)`** is now computed unconditionally per row (was inside the `not has_unrealized` branch). Since `replay_size_at` processes all events up to `ts+60_000` monotonically, calling it every row doesn't change the resulting `replay_size` at any `ts`; it makes the flatten-timestamp capture precise (attributed to the earliest row whose window contains the flatten event), which is what the new `replay_flattened_this_row` logic needs. The old conditional branch still consumes `replay_position_size` correctly (`if replay_ambiguous or replay_position_size > 1e-12`). No double-call.

**Test** `test_coin_hsl_history_replay_resets_current_episode_after_nonpanic_flatten` pins the target behavior: a drawdown episode (upnl −30, realized −20) flattened with `panic_flatten_events: []` (non-panic) and now green → `pnl_reset_timestamp_ms == 180_501` (flatten+1), `halted is False`, `last_stop_event is None`, `last_metrics.tier == "green"`, `drawdown_raw ≈ 0.0`, and `_runtime_forced_modes == {"long": {}, "short": {}}` (no lingering protective mode). Exactly the historical-RED-no-panic-if-now-green outcome.

### Environment caveat
`tests/test_hsl_coin_mode.py` is skipped in my environment (stale `passivbot_rust` extension) → static review + CI. I verified the reset method/closure reuse, the gate composition, and the test assertions by reading; the behavior is consistent with the protective-stop fail-safe model. As a MAX-care coin-HSL replay change, the Hermes half of the two-reviewer gate should confirm at this head.

Net: a well-gated, well-tested coin-HSL episode reset that resolves stale non-panic-flattened RED episodes without touching panic no-restart accounting, fails loud on ambiguous proof, and reuses the established reset path. Approving `5917b31a`.

vigilpbclaw-hash:
PR: #1104
Head SHA: 5917b31a0fb6a836a22c18f6ebdf2940038d681a
Base SHA/branch: 3c597bd94ba28519a220b0d8eeb73b9dc1b3be5f/v8
Scope reviewed: full PR
Verdict: approve

Supersedes: no previous Hermes Continuous Reviewer review was found for this PR.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: the PR narrows coin-HSL history replay so an ordinary, non-panic flat fill terminates the current coin episode and resets current-episode realized-PnL/runtime drawdown before later re-entry, while keeping panic-marker cooldown/no-restart handling separate. That fits the stated partial fable-audit scope: it uses the existing exchange-derived fill replay contract instead of introducing a new state source, preserves conservative behavior when replay is ambiguous, and leaves broader terminal no-restart accounting as future work rather than mixing contracts in this slice.

Evidence reviewed:
- `src/passivbot_hsl.py`: per-coin replay size tracking, non-ambiguous flatten detection, reset timestamp/baseline update, rolling-window reset, and unchanged panic-marker branch.
- `tests/test_hsl_coin_mode.py`: new regression covering increase -> ordinary decrease-to-flat -> re-entry, proving current-episode reset avoids carrying the prior realized loss into live metrics.
- `docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md`: scope/status update matches the implemented partial slice.

Validation run:
- Built/installed `passivbot_rust` in a temporary Python 3.12 venv for this worktree.
- `python -m pytest tests/test_hsl_coin_mode.py::test_coin_hsl_history_replay_resets_current_episode_after_nonpanic_flatten -q` — passed (1 passed).
- `python -m pytest tests/test_hsl_coin_mode.py -q -k 'not test_coin_hsl_check_tp_only_orange_skips_flat_symbols'` — passed (46 passed).
- `python -m pytest tests/test_hsl_coin_mode.py -q` — 46 passed / 1 failed; the failure is `test_coin_hsl_check_tp_only_orange_skips_flat_symbols`, and the same assertion reproduces on clean `origin/v8`, so I am treating it as an unrelated base failure rather than a blocker for this PR.
- `python -m py_compile src/passivbot_hsl.py tests/test_hsl_coin_mode.py` — passed.

PR #1102, opened 2026-07-05
Warn about HSL history reinterpretation

enarjord:
## Summary
- add a dedicated HSL risks doc covering stateless replay, transfers, balance overrides, and config-change reinterpretation
- emit a startup warning when HSL is enabled, pointing to the doc
- add live-config-preflight HSL evidence for the same caveat

## Tests
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m pytest tests/test_live_config_preflight.py tests/test_hsl_coin_mode.py::test_parse_hsl_config_warns_about_history_reinterpretation -q
- /Users/eiriknarjord/repos/passivbot-2/venv/bin/python -m py_compile src/passivbot_hsl.py src/tools/live_config_preflight.py
- git diff --check

PR #1111, opened 2026-07-06
Add fail-closed HSL replay cache loader

enarjord:
## Summary
- add a private HSL replay matrix cache loader that validates manifest metadata and array hashes before returning cached data
- return copied manifest/array data only after validation succeeds; corrupt or mismatched caches raise with existing reason codes
- record the partial HSL replay performance/readiness milestone in the fable-audit plan

## Behavior
- Observability/performance plumbing only; loader remains unwired from live trading decisions
- Cache remains non-authoritative and fail-closed

## Tests
- ./venv/bin/python -m pytest tests/test_hsl_coin_mode.py -k hsl_replay_matrix_cache -q
- ./venv/bin/python -m pytest tests/test_hsl_coin_mode.py -k 'not test_coin_hsl_check_tp_only_orange_skips_flat_symbols' -q
- ./venv/bin/python -m compileall src/passivbot_hsl.py tests/test_hsl_coin_mode.py
- git diff --check

### DIFF
--- a/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
+++ b/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
@@ -852,6 +852,10 @@ Remaining implementation details:
       runs the replay-cache validator, reports valid/invalid cache counts and
       validation reason counts under risk metadata, and emits warning issues
       for caches that must be rebuilt rather than reused.
+      Partial: pure live-HSL cache helpers now include a fail-closed loader
+      that validates manifest metadata and array hashes before returning copied
+      manifest/array data for future replay reuse. The loader remains unwired
+      and non-authoritative.
 - [ ] Python simplification after Rust owns ideal protective orders and unstuck
       orders.
       Python should reconcile ideal vs actual orders, not re-decide trading
--- a/src/passivbot_hsl.py
+++ b/src/passivbot_hsl.py
@@ -756,6 +756,28 @@ def _hsl_replay_cache_validation_reasons(
     return reasons
 
 
+def _load_hsl_replay_matrix_cache(
+    cache_dir: str | os.PathLike[str],
+    *,
+    expected_metadata: dict[str, Any] | None = None,
+) -> tuple[dict[str, Any], dict[str, Any]]:
+    import numpy as np
+
+    reasons = _hsl_replay_cache_validation_reasons(
+        cache_dir,
+        expected_metadata=expected_metadata,
+    )
+    if reasons:
+        raise ValueError("HSL replay cache validation failed: " + ", ".join(reasons))
+    cache_path = Path(cache_dir)
+    manifest = json.loads(
+        (cache_path / _HSL_REPLAY_CACHE_MANIFEST_FILENAME).read_text(encoding="utf-8")
+    )
+    with np.load(cache_path / _HSL_REPLAY_CACHE_MATRIX_FILENAME, allow_pickle=False) as loaded:
+        arrays = {field: loaded[field].copy() for field in _HSL_REPLAY_MATRIX_RAW_FIELDS}
+    return manifest, arrays
+
+
 def _hsl_psides(self) -> tuple[str, str]:
     return ("long", "short")
 
--- a/tests/test_hsl_coin_mode.py
+++ b/tests/test_hsl_coin_mode.py
@@ -260,6 +260,34 @@ def test_hsl_replay_matrix_cache_reports_metadata_mismatch(tmp_path):
     assert reasons == ["metadata_mismatch:config_digest"]
 
 
+def test_hsl_replay_matrix_cache_load_returns_manifest_and_arrays(tmp_path):
+    import numpy as np
+
+    metadata = _hsl_cache_metadata()
+    written_manifest = hsl._write_hsl_replay_matrix_cache(tmp_path, _hsl_cache_rows(), metadata)
+
+    manifest, arrays = hsl._load_hsl_replay_matrix_cache(
+        tmp_path,
+        expected_metadata=metadata,
+    )
+
+    assert manifest == written_manifest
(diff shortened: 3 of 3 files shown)


vigilpbclaw-hash:
PR: #1111
Head SHA: 7a68b99f8520427d2484bac97689ded4fd8c7f45
Base SHA/branch: 3035a28291d205145313770e9a9169acfcc796c0/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this PR adds a private, fail-closed reader for the existing HSL replay matrix cache and deliberately leaves it unwired from live trading. That is the right incremental shape for a future replay reuse path: validation remains centralized in `_hsl_replay_cache_validation_reasons`, manifest metadata and array hashes are checked before any data is returned, and returned arrays are copied out of the `np.load` handle so callers cannot depend on mmap/zipfile lifetime. Because the helper is currently private and non-authoritative, it does not introduce a silent fallback or live risk/order behavior dependency.

Evidence reviewed:
- `src/passivbot_hsl.py` `_load_hsl_replay_matrix_cache`: validates manifest/schema/metadata/array hashes via the existing validator before reading; uses `allow_pickle=False`; returns copied arrays for all raw replay fields.
- `tests/test_hsl_coin_mode.py`: new coverage for successful load, metadata mismatch rejection, and corrupt matrix/hash mismatch rejection.
- `docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md`: milestone note accurately describes the loader as partial/unwired/non-authoritative.

Validation run:
- `git diff --check refs/remotes/origin/v8...HEAD` — passed.
- Dry-run merge of `origin/pr-1111` into current `origin/v8` (`3035a28291d205145313770e9a9169acfcc796c0`) — passed.
- `/Users/nodus/passivbot/venv/bin/python -m pytest tests/test_hsl_coin_mode.py -k hsl_replay_matrix_cache -q` — passed (15 tests).
- `/Users/nodus/passivbot/venv/bin/python -m pytest tests/test_hsl_coin_mode.py -k 'not test_coin_hsl_check_tp_only_orange_skips_flat_symbols' -q` — passed (65 tests).
- `/Users/nodus/passivbot/venv/bin/python -m compileall src/passivbot_hsl.py tests/test_hsl_coin_mode.py` — passed.

enarjord (commented):
**PR:** #1111 — Add fail-closed HSL replay cache loader
**Head SHA:** `7a68b99f8520427d2484bac97689ded4fd8c7f45`
**Base:** v8 (merge-base `3035a282` = current tip — branch is right at HEAD)
**Scope:** `src/passivbot_hsl.py` (+22: `_load_hsl_replay_matrix_cache`), `tests/test_hsl_coin_mode.py` (+43: 3 tests), audit doc (+4). MAX-care area (coin-HSL replay, contract c) but purely additive, unwired scaffolding.
**Verdict:** ✅ **Approve** — no blocking findings; a correct, fail-closed loader that is not wired into the live path (zero behavior change).

This completes the write→validate→load triad for the (still non-authoritative) HSL replay matrix cache. `_load_hsl_replay_matrix_cache(cache_dir, *, expected_metadata=None)` runs `_hsl_replay_cache_validation_reasons` **first** and, if the reason list is non-empty, raises `ValueError("HSL replay cache validation failed: " + reasons)` — returning manifest + copied arrays **only** on a fully clean validation. This is exactly the "fail-safe on missing proof / checkpoints=cache-only" contract: a cache that fails any structural/hash/metadata/value/dtype check is refused, never partially loaded.

**Confirmed unwired.** `git grep` across the PR tree shows `_load_hsl_replay_matrix_cache` has **no caller in `src/`** outside its own definition — only the three new tests invoke it. So there is no change to the live replay/red/panic path and no 9-Goal impact; the audit doc correctly states "The loader remains unwired and non-authoritative."

**The loader is correct and fail-closed:**
- **Validate-before-load ordering.** Reasons are computed before the manifest/npz are opened for return; a non-empty list short-circuits to a raise, so no data from an invalid cache is ever handed back (fail-closed, not fail-open).
- **Returns copies.** Arrays are materialized via `loaded[field].copy()` under the `np.load(...)` context manager, so the returned buffers are independent of the memory-mapped file (no lazy-read surprises or accidental mutation of on-disk state).
- **Metadata gate is threaded through.** `expected_metadata` is passed straight to the validator, so a config/user/mode mismatch (`metadata_mismatch:<key>`) rejects the load — preserving the #1102 history-reinterpretation caveat (a cache built under a different config/signal_mode is not silently reused).

**Verified by direct exercise** (main-clone Rust `.so` + numpy; `test_hsl_coin_mode.py` skips via pytest on the stale extension): a valid write→load returns `manifest == written_manifest`, all six raw-field arrays, and the expected `ts`; loading with a mismatched `config_digest` raises `ValueError` containing `metadata_mismatch:config_digest`; tampering the on-disk `pnl` array then loading raises `ValueError` containing `array_hash_mismatch:pnl` (stale/tampered data refused). Matches the three new tests exactly.

**Value-safety.** N/A as a leak surface — this is an internal loader returning in-memory manifest/arrays for future replay reuse, not a shareable smoke-report projection. No magnitudes are emitted anywhere new.

### Environment caveat
`test_hsl_coin_mode.py` isn't runnable via pytest here (module-level skip on the stale `passivbot_rust` extension) → I verified the loader by direct invocation; CI runs the suite. As a MAX-care coin-HSL change (even unwired), the Hermes half of the gate should confirm at this head and confirm the intended wiring plan (and, when the loader is eventually wired, that a load failure triggers a rebuild-from-authoritative-inputs rather than a fall-through).

Net: a small, correct, fail-closed loader that validates before returning and hands back independent copies — completing the cache write/validate/load scaffolding while remaining unwired from the live path (no behavior change), pinned by tests and locally verified. Approving `7a68b99f`.

PR #1125, opened 2026-07-06
Extract HSL fill-event normalization into reusable methods

enarjord:
Behavior-neutral refactor ahead of the replay-cache reuse wiring (parallel to #1124; different files, no shared hunks). The trust-critical fill normalization lived in closures inside `get_balance_equity_history`; the reuse path must apply the identical normalization to post-watermark gap fills, and duplicating it would fork the trust boundary.

Scope (`src/passivbot.py` + one test):

- `_hsl_extract_fill_events`, `_hsl_normalize_fill_symbol`, `_hsl_fill_action`, `_hsl_fill_safe_float` are now `Passivbot` methods with bodies moved **verbatim** from the closures. The remaining in-function uses of `_safe_float`/`_normalize_symbol` (progress payloads, position state, price handling) delegate to the methods via local aliases, so there is exactly one implementation.
- Direct unit coverage of the extraction contract (previously only testable through the full replay): exchange-id symbol normalization via `get_symbol_id_inv`, qty fallback chain (`qty_signed` → `qty`/`amount`/`size`/`contracts`), price fallback chain (`price` → `avgPrice`/`average`/… → `info`), signed-qty and side-based action inference (short buy → decrease), unsigned qty storage, `c_mult` stamping, and the skip rules (missing timestamp/symbol, zero qty/price).

Behavior neutrality evidence: the 200+ `get_balance_equity_history` tests in `test_passivbot_balance_split.py` (including the matrix-collection and pacing tests) pass unchanged — 203 passed; `tests/test_hsl_coin_mode.py` + `tests/test_unstucking_safeguards.py` — 204 passed. Full suite: only the 3 known pre-existing out-of-scope failures (see #1116). `git diff --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1124, opened 2026-07-06
Persist panic flatten markers with the HSL account series

enarjord:
Third and final precursor to the replay-cache reuse gate. Wiring recon found that `_equity_hard_stop_initialize_coin_from_history` consumes `panic_flatten_events` (derived during the full replay's fill accounting via after-psize flat detection) for cooldown/no-restart reconstruction — a cache-fed replay would silently lose in-window markers. They are raw exchange-derived facts (timestamp, minute, pside, symbol), so they belong in the cache, still write-only.

Scope:

- **Cache schema v4**: account-series manifests always carry `panic_flatten_events` (possibly empty), normalized fail-loud — required fields, `long`/`short` pside, non-empty symbol, minute-grid alignment, series-span bounds, ascending timestamps. Pair manifests must not carry them (`panic_flatten_events` are account-scoped facts; the writer rejects the combination).
- **Validator**: `panic_events_missing` (a v4 account manifest without the field is invalid — this matters because "no field" must never be read as "no panics"), `panic_events_invalid` (corrupted entries), `panic_events_wrong_kind` (markers on a pair manifest). Schema bump v3→v4 rejects all prior caches rather than reinterpreting them; nothing reads caches yet, so only doctor reporting changes.
- **Persist helper** threads `history["panic_flatten_events"]` through to the account write; the end-to-end persist round-trip test verifies the markers land in the manifest and survive the fail-closed loader.
- **Plan doc**: records the strengthened centralization guiding contract (user decision 2026-07-06): any HSL/risk/unstuck contract shared by live and backtest must be implemented in Rust and consumed by both paths unless the implementing PR explicitly documents why centralization is infeasible or materially worse; Python-owned live HSL code is limited to exchange data gathering, cache/replay matrix construction, diagnostics, and reconciliation. Plus the tracker note for this slice.

Validation:
- `venv/bin/python -m pytest tests/test_hsl_coin_mode.py tests/test_passivbot_balance_split.py tests/test_cache_integrity_doctor.py tests/test_unstucking_safeguards.py` — 425 passed
- New tests: marker round-trip through writer/validator/loader with tamper detection (stripped field, corrupted entry); fail-loud contracts (markers on pair kind, off-grid minute, out-of-span minute, wrong-kind manifest tampering); persist wiring carries the history's markers.
- Full suite: only the 3 known pre-existing out-of-scope failures (pymoo sampling, 2 bitget UTA stubs; see #1116).
- `git diff --check` — clean

[...]

PR #1123, opened 2026-07-06
Add pure HSL cache watermark-extension helpers

enarjord:
Second precursor to the replay-cache reuse gate (after the #1121 parity trust boundary): pure, unwired helpers that extend a valid cache from its watermark to now, so the reuse slice only needs to fetch fills/candles for the gap instead of the full lookback.

Scope (`src/passivbot_hsl.py` + tests, no wiring):

- `_hsl_replay_rows_from_arrays` — validated, kind-aware inverse of the arrays builders, used to compose cached arrays with extension rows.
- `_hsl_replay_extend_pair_rows` — extends a cached pair matrix using the pair's post-watermark extracted fills plus candle closes. Position bookkeeping mirrors the authoritative `_apply_event` exactly (weighted-average entry price on increase; size shrink with price zeroing on flat), prices carry forward through close gaps, per-minute `pnl` is `pnl + fee` like the authoritative accounting.
- `_hsl_replay_extend_account_rows` — extends the account PnL series with per-minute `pnl + fee` sums over all post-watermark fills.
- **Fail-loud contracts**: a fill inside the cached window (double-count hazard), a fill beyond the extension end, a fill from another pair, an extension end before the watermark, or invalid cached arrays all raise — the future reuse gate must treat any of these as cache-reject-and-fall-back, never as repairable.
- **Slice-and-extend parity**: a five-minute scenario (deep dip, partial close, post-watermark re-entry) runs through the **real** collection; cutting the arrays at a watermark and extending with the post-watermark fills/closes rebuilds arrays equal to the full-run arrays to `atol=1e-12` for both series. Combined with #1121's synthesis parity, the full reuse path (load → extend → synthesize → replay) is now parity-proven piecewise before any wiring.

Validation:
- `venv/bin/python -m pytest tests/test_hsl_coin_mode.py tests/test_passivbot_balance_split.py tests/test_unstucking_safeguards.py` — 406 passed
- Full suite: only the 3 known pre-existing out-of-scope failures (pymoo sampling, 2 bitget UTA stubs; see #1116).
- `git diff --check` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1121, opened 2026-07-06
Add HSL replay-cache parity trust boundary

enarjord:
The parity slice the plan requires before any cache reuse ("read/extend/reuse behavior only after trust-boundary tests are strong"). Introduces the pure, still-unwired conversion the read slice will consume, and proves — with product code on both sides — that cache-derived replay is indistinguishable from the authoritative one.

Scope:

- `src/passivbot_hsl.py`: new pure `_hsl_replay_timeline_rows_from_cache(pair_arrays, account_arrays, *, current_balance)` synthesizes coin-replay timeline rows from persisted arrays. Balance is anchored so the final minute equals current balance (mirroring the authoritative baseline back-computation); per-pair realized values are cumulative from matrix start — the replay windowing consumes differences, so a constant anchor offset is harmless by contract, and the docstring says so. Fail-loud on: invalid arrays (kind-aware value checks), non-contiguous timestamps, pair spans outside the account span, minute-grid misalignment, and any synthesized balance ≤ 0 (the coin metric layer requires balance > 0).
- Parity test 1 (row level): a two-fill scenario (buy, deep dip to 60, partial close, recovery) runs through the **real** `get_balance_equity_history` collection; rows synthesized from the returned matrices/account series match the authoritative timeline value-for-value (balance, scoped realized, upnl) wherever the pair exists, and any earlier synthesized-only entries (pair rows may start at the first candle, before the first fill) are exactly neutral.
- Parity test 2 (state level): the **real** `_equity_hard_stop_initialize_coin_from_history` replays both timelines on two coin bots with a per-sample recorder; the aligned metric sequences (timestamp, tier, drawdown_raw/score, balance, upnl) are **identical across an orange tier transition**, extra leading synthesized samples must be green/zero, and final metrics + halted/cooldown state are equal.
- Fail-loud tests for bad inputs.

Nothing is wired: `git grep` shows the new helper has no caller outside tests. The read slice (next) will feed this synthesizer through the existing initializer loop, gated on proven manifests + fresh coverage proof + watermark extension, per the gate notes on #1118/#1119.

Validation:
- `venv/bin/python -m pytest tests/test_hsl_coin_mode.py tests/test_passivbot_balance_split.py tests/test_unstucking_safeguards.py` — 404 passed
- Full suite: only the 3 known pre-existing out-of-scope failures (pymoo sampling, 2 bitget UTA stubs; see #1116).
- `git diff --check` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1126, opened 2026-07-06
Wire HSL replay cache reuse into live coin startup

enarjord:
The first behavior-bearing slice of the replay-cache work (B2.1b), composing the six merged precursors: metadata/digests (#1115/#1118), write persistence (#1117), account series (#1119), synthesis parity (#1121), extension parity (#1123), persisted panic markers (#1124), and shared fill extraction (#1125).

**What changes**: live coin-mode HSL startup calls `_equity_hard_stop_try_reuse_replay_cache` before `get_balance_equity_history`. On success, the replay history is reconstructed from the persisted caches plus only the watermark→now gap; on any gate failure or unexpected error, it falls back to the authoritative full replay (shutdown cancellation still propagates). The reused history flows through the **same** downstream code: the existing replay loop via the parity-proven synthesizer, the persisted panic markers from the account manifest, and the unchanged re-persist path that refreshes the cache.

**Reuse gates (each individually tested to fall back)**:
1. Write-time `fill_coverage_proven=true` in the manifest **and** a fresh load-time `_pnl_history_coverage_status` proof — both required, per the #1118 contract.
2. Expected-metadata identity through the fail-closed loader: coverage window/scope come from the manifest itself (tautological compare — the reader cannot predict write-time windows) while exchange/user/market_type/config-digest/signal-mode are enforced strictly, `fill_coverage_proven` must equal `true`, and scope must not be `unknown`.
3. Account and every held-pair watermark must agree.
4. Gap fills come from the pnls manager through the shared #1125 extraction; gap candles are fetched only for the watermark gap. A panic fill inside the gap forces the full replay (marker derivation needs flat-detection state the reuse path does not reconstruct — deliberately conservative).
5. The extended series' final position must reconcile with current exchange positions within qty-step tolerance.

**Statelessness**: unchanged — the cache accelerates, never decides. A fresh VPS (no cache) takes the full-replay path and reaches identical state; the end-to-end test proves the two paths are state-identical.

**Evidence**:
- End-to-end identity test: boot 1 full-replays a prefix window and persists; boot 2 reuses over the extended window with `get_balance_equity_history` replaced by a raiser (proving the full fetch is skipped) and reaches coin state identical to a control full replay (tier, drawdowns, balance, upnl to 1e-9, halted/cooldown equal).
- Rejection tests: cache miss, HSL config change (digest rotation), irreconcilable positions, panic fill in the gap, unproven load-time coverage — each returns None and falls back.
- Fixed along the way: `_try_load_hsl_replay_matrix_cache` was never bound on `Passivbot` (unwired scaffolding); it and the new reuse method are now bound and pinned in the AST test + fake-bot bind list.
- `HSL_REPLAY_COMPLETED` now reports `cache_reused` for boot-timing evidence (plan requirement: prove the checkpoint actually speeds boot).
- CHANGELOG entry (user-facing behavior change) + tracker update.
- 429 passed across the six HSL/replay/doctor test files; full suite shows only the 3 known pre-existing out-of-scope failures (see #1116). `git diff --check` clean.

**Known conservative limitations (intentional, documented here)**: reuse requires all held pairs cached with agreeing watermarks; gap panic fills, inverse markets, and non-coin signal modes always take the full replay. Flat-pair cooldown reconstruction still relies on the persisted markers plus fill-derived contracts exactly as the full replay does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Live coin-mode HSL startup can now reuse its persisted replay cache: when
+  the cached series pass every trust gate (proven fill coverage at write and
+  load time, config digest identity, watermark agreement, gap extension from
+  exchange fills/candles, and current-position reconciliation), the bot
+  replays from the cache plus the gap instead of re-fetching the full
+  lookback. Any gate failure falls back to the full exchange-derived replay;
+  the cache never becomes authoritative trading state, and a fresh VPS
+  reconstructs identical decisions.
 - HSL-enabled startup and live-config preflight now surface a history
   reinterpretation caveat and point operators to a dedicated HSL risks doc for
   deposits, withdrawals, balance overrides, and HSL config changes.
--- a/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
+++ b/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
@@ -996,6 +996,16 @@ Remaining implementation details:
       cooldown/no-restart evidence. Markers are validated fail-loud (grid
       alignment, series-span bounds, ascending order, account-kind only) and
       tamper-checked (missing/invalid/wrong-kind reasons).
+      Implemented: live coin-mode startup now attempts cache reuse before the
+      full replay. Gates: write-time proven coverage recorded in the manifest
+      plus a fresh load-time pnls-manager coverage proof, config-digest
+      identity, account/pair watermark agreement, gap extension from
+      exchange fills/candles (panic fills inside the gap force full replay),
+      and current-position reconciliation within qty-step tolerance. Any
+      rejection or unexpected error falls back to the authoritative full
+      replay; the completed event now reports `cache_reused` alongside phase
+      timings. End-to-end test proves the cache-fed boot reaches state
+      identical to the full replay with the full fetch provably skipped.
 - [ ] Python simplification after Rust owns ideal protective orders and unstuck
       orders.
       Python should reconcile ideal vs actual orders, not re-decide trading
--- a/src/passivbot.py
+++ b/src/passivbot.py
@@ -1780,6 +1780,10 @@ class Passivbot:
     _equity_hard_stop_persist_replay_matrices = (
         pb_hsl._equity_hard_stop_persist_replay_matrices
     )
+    _try_load_hsl_replay_matrix_cache = pb_hsl._try_load_hsl_replay_matrix_cache
+    _equity_hard_stop_try_reuse_replay_cache = (
+        pb_hsl._equity_hard_stop_try_reuse_replay_cache
+    )
     _equity_hard_stop_maybe_emit_raw_red_pending = (
         pb_hsl._equity_hard_stop_maybe_emit_raw_red_pending
     )
--- a/src/passivbot_hsl.py
+++ b/src/passivbot_hsl.py
@@ -1609,6 +1609,282 @@ def _try_load_hsl_replay_matrix_cache(
     return manifest, arrays
 
 
+def _hsl_replay_cache_read_manifest_window(
+    cache_dir: str | os.PathLike[str],
+) -> dict[str, Any] | None:
+    """Best-effort read of a manifest's recorded coverage window and scope.
+
+    The reader cannot predict the write-time coverage window, so expected
+    metadata takes those fields from the manifest itself (tautological
+    comparison) while identity fields and the coverage proof stay strict.
+    Returns None when no readable manifest exists (a cache miss).
+    """
+    manifest_path = Path(cache_dir) / _HSL_REPLAY_CACHE_MANIFEST_FILENAME
+    if not manifest_path.exists():
+        return None
+    try:
+        manifest = json.loads(manifest_path.read_text(encoding="utf-8"))
+        metadata = manifest.get("metadata") or {}
+        return {
+            "fill_covered_start_ms": int(metadata["fill_covered_start_ms"]),
+            "fill_covered_end_ms": int(metadata["fill_covered_end_ms"]),
+            "fill_history_scope": str(metadata["fill_history_scope"]),
+            "candle_covered_start_ms": int(metadata["candle_covered_start_ms"]),
+            "candle_covered_end_ms": int(metadata["candle_covered_end_ms"]),
+        }
+    except Exception:
+        # Unreadable manifests are handled (and evented) by the validator.
+        return {}
+
+
+async def _equity_hard_stop_try_reuse_replay_cache(
+    self, now_ms: int
+) -> dict[str, Any] | None:
+    """Attempt to reconstruct the coin-replay history from persisted caches.
+
+    Returns a history payload equivalent to `get_balance_equity_history()`
+    output for the coin replay, or None when the cache must not be used, in
+    which case the caller falls back to the full exchange-derived replay.
+    Every rejection is observable through the cache hit/miss/rejected events
+    emitted by the loader or a warning here. The cache never becomes
+    authoritative: reuse requires a proven write-time fill-coverage flag, a
+    fresh load-time coverage proof, digest identity, watermark agreement,
+    gap extension from exchange fills/candles, and current-position
+    compatibility; any failure falls back.
+    """
+    if self._equity_hard_stop_signal_mode() != "coin":
+        return None
+    if bool(getattr(self, "inverse", False)):
+        return None
+    manager = getattr(self, "_pnls_manager", None)
+    if manager is None:
+        return None
+    held_pairs: list[tuple[str, str]] = []
+    for symbol, slots in (self.positions or {}).items():
+        if not isinstance(slots, dict):
+            continue
+        for pside in self._hsl_psides():
+            if not self._equity_hard_stop_coin_active_pside(pside):
+                continue
+            if self._equity_hard_stop_has_open_position_symbol(pside, str(symbol)):
+                if not self._equity_hard_stop_symbol_supported_for_coin_replay(
+                    str(symbol)
+                ):
+                    return None
+                held_pairs.append((pside, str(symbol)))
+    if not held_pairs:
+        return None
+
+    lookback = parse_pnls_max_lookback_days(
+        self.live_value("pnls_max_lookback_days"),
+        field_name="live.pnls_max_lookback_days",
+    )
+    lookback_start = lookback.balance_history_start_ms(int(now_ms))
+    coverage_status = self._pnl_history_coverage_status(
+        start_ms=None if lookback_start is None else int(lookback_start),
+        end_ms=int(now_ms),
+        lookback=lookback,
+    )
+    if not bool(coverage_status.get("ready", False)):
+        logging.info(
+            "[risk] HSL replay cache reuse skipped: fill coverage not proven at load "
+            "| reason=%s",
+            str(coverage_status.get("reason", "unknown")),
+        )
+        return None
+
+    def expected_from_window(window: dict[str, Any], *, account: bool, pside: str = "", symbol: str = ""):
+        if window.get("fill_history_scope") not in ("window", "all"):
+            return None
+        kwargs = {
+            "fill_covered_start_ms": window["fill_covered_start_ms"],
+            "fill_covered_end_ms": window["fill_covered_end_ms"],
+            "fill_history_scope": window["fill_history_scope"],
+            # Strict: only caches written from proven fills are reusable.
+            "fill_coverage_proven": True,
+            "candle_covered_start_ms": window["candle_covered_start_ms"],
+            "candle_covered_end_ms": window["candle_covered_end_ms"],
+        }
+        if account:
+            return self._hsl_replay_cache_account_expected_metadata(**kwargs)
+        return self._hsl_replay_cache_expected_metadata(pside, symbol, **kwargs)
+
+    account_dir = self._hsl_replay_cache_account_series_dir()
+    account_window = _hsl_replay_cache_read_manifest_window(account_dir)
+    if account_window is None or not account_window:
+        return None
+    account_expected = expected_from_window(account_window, account=True)
+    if account_expected is None:
+        return None
+    account_loaded = self._try_load_hsl_replay_matrix_cache(
+        account_dir,
+        expected_metadata=account_expected,
+        pside=_HSL_REPLAY_CACHE_ACCOUNT_PSIDE,
+        symbol=_HSL_REPLAY_CACHE_ACCOUNT_SYMBOL,
+    )
+    if account_loaded is None:
+        return None
+    account_manifest, account_arrays = account_loaded
+    watermark_ts = int(account_manifest["end_ts_ms"])
+
+    pair_arrays_by_pair: dict[tuple[str, str], dict[str, Any]] = {}
+    for pside, symbol in held_pairs:
+        pair_dir = self._hsl_replay_cache_dir(pside, symbol)
+        pair_window = _hsl_replay_cache_read_manifest_window(pair_dir)
+        if pair_window is None or not pair_window:
+            return None
+        pair_expected = expected_from_window(
+            pair_window, account=False, pside=pside, symbol=symbol
+        )
+        if pair_expected is None:
+            return None
+        pair_loaded = self._try_load_hsl_replay_matrix_cache(
+            pair_dir,
+            expected_metadata=pair_expected,
+            pside=pside,
+            symbol=symbol,
+        )
+        if pair_loaded is None:
+            return None
+        pair_manifest, pair_arrays = pair_loaded
+        if int(pair_manifest["end_ts_ms"]) != watermark_ts:
+            logging.warning(
+                "[risk] HSL replay cache reuse rejected: pair %s:%s watermark %s "
+                "disagrees with account watermark %s",
+                pside,
+                symbol,
+                pair_manifest["end_ts_ms"],
+                watermark_ts,
+            )
+            return None
+        pair_arrays_by_pair[(pside, symbol)] = pair_arrays
+
+    raw_fills = [event.to_dict() for event in manager.get_events()]
+    extracted = self._hsl_extract_fill_events(raw_fills)
+    gap_floor = watermark_ts + _HSL_REPLAY_MATRIX_INTERVAL_MS
+    gap_fills = [event for event in extracted if int(event["timestamp"]) >= gap_floor]
+    if any(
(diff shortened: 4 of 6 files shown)


enarjord:
@vigilpbclaw-hash Hermes, Claude, Codex: requesting review of this PR (head 64a4deaed7570546464361afe630f18b873c5ee6).

This is the MAX-care behavior-bearing slice the whole #1115–#1125 stack built toward, so full scrutiny is warranted. The claims that matter most:
1. **Hit-path exactness**: a cache-fed boot must be state-identical to the authoritative replay — the end-to-end test proves it for the covered scenario; if you can construct a scenario where they diverge without a gate rejecting, that's the critical finding.
2. **Fail-safe completeness**: every gate failure and every unexpected exception must land on the full replay (never abort startup, never proceed on partial trust). Check the guard placement around the reuse call and the CancelledError re-raise.
3. **No authoritative-state leak**: the reused data is validated cache + fresh exchange gap; verify nothing skips the load-time coverage proof or accepts an unproven/absent manifest field.
4. Per the plan's centralization contract: this slice is Python because it is exchange-I/O + cache/matrix construction + reconciliation — the threshold/tier decisions remain in the shared Rust runtime consumed by the replay loop.

vigilpbclaw-hash:
PR: #1126
Head SHA: 64a4deaed7570546464361afe630f18b873c5ee6
Base SHA/branch: 82eccc83c956ba75d976efdcb43f1904ef7a6be8/v8
Scope reviewed: full PR
Verdict: approve

Findings: none.

Intent/architecture pass: this PR wires previously persisted HSL replay matrices/account series into live coin-mode startup as an accelerator, while keeping the full replay as the authoritative fallback. The design is appropriately conservative for trading/risk code: reuse is limited to coin mode/non-inverse/current held pairs, requires write-time and load-time fill coverage proof, strict metadata/config identity, account/pair watermark agreement, gap reconstruction from canonical fill extraction plus candles, panic-gap fallback, and final position reconciliation before the synthesized timeline enters the existing coin replay loop. That matches the stated “cache accelerates, never decides” intent and avoids adding a parallel HSL decision path.

Blocker:
- None.

Important:
- None.

Minor:
- None.

Validation performed:
- `git diff --check origin/v8...HEAD` — clean.
- Dry-run merge into current `origin/v8` — automatic merge succeeded.
- `python3 -m py_compile src/passivbot.py src/passivbot_hsl.py tests/test_hsl_coin_mode.py tests/test_passivbot_hsl_bindings.py` — passed.
- Targeted pytest command for the new HSL cache reuse tests and binding test could not run in this review environment because the available Python environment lacks `ccxt` (`ModuleNotFoundError: No module named 'ccxt'`). I treated that as an environment limitation rather than a branch blocker and relied on static review plus compile/merge checks.

Notes:
- The Passivbot-specific review rubric files referenced by the review workflow were not present in this checkout under `references/`; I applied the relevant HSL/live-risk/fill-accounting checks manually from the diff and surrounding code.

enarjord:
Codex review for PR #1126 at `64a4deaed7570546464361afe630f18b873c5ee6`: **needs fix**.

Finding:

- **P1: cache reuse can surface incomplete flat-pair cooldown history instead of falling back.** `_equity_hard_stop_try_reuse_replay_cache()` builds `held_pairs` only from currently open positions (`src/passivbot_hsl.py:1662-1676`), loads/extends pair matrices only for those pairs (`src/passivbot_hsl.py:1730-1760`, `1797-1815`), then returns every account-manifest `panic_flatten_events` marker (`src/passivbot_hsl.py:1856-1861`). The coin replay initializer treats each supported panic marker as a required replay pair (`src/passivbot_hsl.py:4393-4413`) and later requires per-coin timeline values from `required_start_ts` onward (`src/passivbot_hsl.py:4575-4657`). A supported flat coin in active HSL cooldown can therefore be present in the account cache markers but absent from the cache-derived timeline, which raises during startup instead of using the authoritative full replay. This violates the PR's fail-safe claim: a cache accelerator must not abort startup or lose flat-pair cooldown reconstruction. The narrow fix can be conservative: if any supported panic marker pair is not covered by a loaded pair matrix, reject reuse and fall back to full replay; or load/extend all panic-marker pairs too. Add a regression where one current held pair is reusable while another supported flat coin has a cached panic marker/cooldown, and assert either full replay is used or the cached replay reconstructs the flat coin cooldown.

Evidence:

- Focused tests passed: `compileall -q src/passivbot.py src/passivbot_hsl.py tests/test_hsl_coin_mode.py tests/test_passivbot_hsl_bindings.py`
- Focused pytest passed: `tests/test_passivbot_hsl_bindings.py`, `test_hsl_cache_synthesized_rows_match_authoritative_timeline`, `test_hsl_cache_synthesized_replay_reaches_identical_coin_state`, `test_hsl_cache_extension_matches_full_rebuild`, `test_hsl_cache_extension_fail_loud_contracts`, `test_hsl_cache_reuse_reaches_state_identical_to_full_replay`, `test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache`, `test_hsl_cache_synthesized_rows_fail_loud_on_bad_inputs`, `test_coin_hsl_history_replay_persists_replay_matrices` (`9 passed`).
- Reviewer-only reproducer in `/private/tmp/test_pr1126_flat_marker_reproducer.py`: after making the flat marker symbol supported, the test confirms `await bot._equity_hard_stop_initialize_coin_from_history()` raises `ValueError: realized_pnl_by_coin_pside` when the account manifest contains a panic marker for a non-held supported symbol and only the current held pair has a matrix. Expected behavior is reuse rejection plus full replay, not startup abort.

enarjord (commented):
**PR:** #1126 — Wire HSL replay cache reuse into live coin startup
**Head SHA:** `64a4deaed7570546464361afe630f18b873c5ee6`
**Base:** v8 (merge-base `82eccc83` = current tip — branch is right at HEAD)
**Scope:** `src/passivbot_hsl.py` (+310: `_equity_hard_stop_try_reuse_replay_cache` + `_hsl_replay_cache_read_manifest_window`/`_hsl_reuse_group_matrices`, and the coin-init wiring), `src/passivbot.py` (+4 bindings), `tests/test_hsl_coin_mode.py` (+398), CHANGELOG + audit doc. **This is the signal-bearing REUSE PR** — the first that reads the replay cache into the live coin red/panic path (contract c, MAX-care).
**Verdict:** ✅ **Approve** — no blocking findings; a rigorously fail-safe, parity-proven wiring where a cache HIT yields identical coin HSL state to the authoritative replay and every doubt falls back to the full exchange-derived replay.

`_equity_hard_stop_try_reuse_replay_cache(now_ms)` reconstructs the coin-replay history from the persisted caches and returns a payload equivalent to `get_balance_equity_history()`, or `None` to force the full replay. It is composed with all the trust-boundary layers built across #1108–#1125.

**The fallback is unconditional and correct.** At the call site the reuse is wrapped in `try/except`: `CancelledError` re-raises; any other exception logs and sets `history = None`. Then `if history is not None: use cache else: history = await self.get_balance_equity_history(...)` — the authoritative full replay. So reuse is a pure accelerator that can never abort startup or degrade the signal; the same downstream red/panic/cooldown logic consumes whichever history is produced. Startup-blocking protective readiness is preserved.

**The gate returns `None` on every doubt (→ full replay):**
- Not coin mode / inverse / no `_pnls_manager` / no held pairs / unsupported held symbol.
- **Fresh load-time coverage proof:** `_pnl_history_coverage_status(...)` must be `ready`, else skip. (#1118 proof re-checked at load, not just write.)
- **Strict write-time proof:** `expected_from_window` requires `fill_history_scope ∈ {window, all}` and forces `fill_coverage_proven=True` — only caches written from proven fills are eligible.
- **Validation + digest identity:** account and every held-pair cache go through `_try_load_hsl_replay_matrix_cache(expected_metadata=…)` (hash/schema/dtype/value + config_digest + coverage metadata); any rejection → `None` (and emits the value-safe hit/miss/rejected event).
- **Watermark agreement:** each pair's `end_ts_ms` must equal the account watermark, else reject.
- **Panic-in-extension-gap → fall back:** any post-watermark fill with a `panic` order type is not reconstructed here (needs flat-detection state), so it defers to the authoritative replay.
- **Extension via proven helpers:** `_hsl_replay_extend_pair_rows`/`_hsl_replay_extend_account_rows` (#1123, proven bit-equal to a full rebuild) extend from the watermark using post-watermark exchange fills + 1m candle closes.
- **Current-position reconciliation (key completeness guard):** the extended series' final `psize` must reconcile with the live exchange position within `qty_step` tolerance, else reject. A missing or extra post-watermark fill makes the reconstruction diverge from the real position → fallback. This is what backs the "extension-window completeness" concern.
- Timeline is then built by `_hsl_replay_timeline_rows_from_cache` (#1121, proven row-for-row equal to the authoritative timeline), and the persisted panic-flatten markers (#1124) are carried through.

**Parity is proven end-to-end, not just per-helper.** I ran the two decisive tests at this head (main-clone Rust `.so`): `test_hsl_cache_reuse_reaches_state_identical_to_full_replay` — **passes**, so the cache-reuse path reaches *identical coin HSL state* to the full replay — and `test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache` — **passes**, exercising the fallback. That end-to-end identical-state test is the strongest possible evidence for a signal-bearing reuse: on a hit the red/panic/tier outcome is the same as if the cache didn't exist.

**Value-safety.** Reuse surfaces observability through the existing value-safe `hsl.replay.cache` hit/miss/rejected events (reason codes/counts, no magnitudes) plus local `logging` (the position-reconcile warning prints position sizes, but that's a local log, not a shareable smoke-report surface — consistent with existing HSL debug logging). No new shareable projection reads risk magnitudes. The CHANGELOG/audit-doc entries describe the reuse as an accelerator that stays non-authoritative and falls back — faithful, no overclaim.

**9-Goal.** Exact red/panic + parity: identical-state test. Fail-safe on missing proof: unproven/missing/incompatible → full replay. Reuse panic supervisor: panic markers persisted and panic-in-gap defers to authoritative. Checkpoints=cache-only: the cache never becomes authoritative (any failure re-derives from exchange). Held protective readiness: fallback never aborts startup.

### Environment / gate note
`test_hsl_coin_mode.py` module-skips under pytest, so I ran the reuse tests individually against the compiled `.so`; CI runs the suite. As the culminating MAX-care coin-HSL change, the Hermes half of the two-reviewer gate should confirm at this head — especially the identical-state parity on CI and, ideally, a soak that exercises real persisted caches (write on one run, reuse on the next) plus the panic-in-gap and position-reconcile fallbacks under live-like fills.

Net: the reuse wiring is a pure, fail-safe accelerator — a cache hit is proven to yield the identical coin HSL state as the authoritative replay, and every unproven/mismatched/incomplete condition (coverage, digest, watermark, panic-in-gap, position reconciliation, or any exception) falls back to the full exchange-derived replay. The whole #1108–#1126 pipeline lands behavior-equivalent-on-hit and behavior-unchanged-on-miss. Approving `64a4deae`.

enarjord:
Pushed 3ecfa33a addressing Codex's P1:

- `_equity_hard_stop_try_reuse_replay_cache` now rejects reuse up front when any account-manifest panic marker targets a pair that is supported for coin replay but not covered by a loaded held-pair matrix — the authoritative full replay then reconstructs that flat-pair cooldown, instead of the synthesized timeline aborting startup mid-replay on the required-pair check. Markers for unsupported symbols remain ignored, matching the full replay's own handling.
- Regression per your spec: a cached marker for a supported flat coin (held pair BTC reusable, flat ETH marker in the account manifest) rejects reuse; flipping the marker symbol to unsupported leaves reuse allowed; the remaining gate checks re-verify against a marker-free cache.

Validation: `venv/bin/python -m pytest tests/test_hsl_coin_mode.py tests/test_unstucking_safeguards.py tests/test_passivbot_balance_split.py tests/test_passivbot_hsl_bindings.py` — all passed; `git diff --check` clean.

@vigilpbclaw-hash Hermes, Claude, Codex: requesting delta re-review of 64a4deae..3ecfa33a.

enarjord:
Codex delta re-review for PR #1126 at `3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641`: **approve**.

Findings: none.

The delta `64a4deaed7570546464361afe630f18b873c5ee6..3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641` addresses my P1. `_equity_hard_stop_try_reuse_replay_cache()` now rejects reuse before synthesis when the account cache has a panic marker for a supported pair that is not among loaded held-pair matrices, so flat-pair cooldown reconstruction falls back to the authoritative full replay instead of entering coin replay with missing per-coin timeline fields. Unsupported-marker behavior remains aligned with full replay's unsupported historical-symbol skip.

Validation:

- `git diff --check 64a4deaed7570546464361afe630f18b873c5ee6..HEAD` — pass.
- `compileall -q src/passivbot.py src/passivbot_hsl.py tests/test_hsl_coin_mode.py tests/test_passivbot_hsl_bindings.py` — pass.
- Focused pytest — pass, `9 passed`:
  - `tests/test_passivbot_hsl_bindings.py`
  - `test_hsl_cache_synthesized_rows_match_authoritative_timeline`
  - `test_hsl_cache_synthesized_replay_reaches_identical_coin_state`
  - `test_hsl_cache_extension_matches_full_rebuild`
  - `test_hsl_cache_extension_fail_loud_contracts`
  - `test_hsl_cache_reuse_reaches_state_identical_to_full_replay`
  - `test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache`
  - `test_hsl_cache_synthesized_rows_fail_loud_on_bad_inputs`
  - `test_coin_hsl_history_replay_persists_replay_matrices`
- Regression-only rerun: `test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache` — pass.

Codex is green for this head. Merge gate still needs fresh Hermes + Claude green reviews for `3ecfa33a` plus green CI.

vigilpbclaw-hash:
PR: #1126
Head SHA: 3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641
Base SHA/branch: 82eccc83c956ba75d976efdcb43f1904ef7a6be8/v8
Scope reviewed: delta 64a4deaed7570546464361afe630f18b873c5ee6..3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641
Verdict: approve

Previous Hermes review for 64a4deaed7570546464361afe630f18b873c5ee6 is superseded by this head SHA.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: this delta adds a fail-closed guard for cached flat-pair panic markers before HSL replay-cache reuse. That matches the PR's conservative cache-accelerates-never-decides contract: when reuse cannot reconstruct flat-pair cooldown state from loaded held-pair matrices, it falls back to authoritative full replay rather than trying to synthesize missing per-coin data. The guard is localized to reuse eligibility and does not alter order/risk actions directly.

Evidence reviewed:
- `src/passivbot_hsl.py:_equity_hard_stop_try_reuse_replay_cache` now rejects reuse when a supported `panic_flatten_events` marker is outside the held-pair matrix set, while ignoring unsupported symbols consistently with the full replay's support filter.
- `tests/test_hsl_coin_mode.py::test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache` adds coverage for supported flat-marker rejection and unsupported-marker allowance.

Validation run:
- `git diff --check 64a4deaed7570546464361afe630f18b873c5ee6..3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641` — passed.
- `python3 -m compileall -q src/passivbot_hsl.py tests/test_hsl_coin_mode.py` — passed.
- Targeted pytest could not be executed in this review environment because `passivbot_rust` fails to import (`symbol not found in flat namespace '_PyErr_GetRaisedException'`), so the dynamic test was skipped by module import rather than exercised here.

enarjord (commented):
**Reviewer:** Claude (automated continuous reviewer) — SHA update
**PR:** #1126 — Wire HSL replay cache reuse into live coin startup
**Head SHA:** `3ecfa33ad22f80ce6eeb72214ce82e9bd62fb641` (amended from previously-reviewed `64a4deae`)
**Base:** v8 (merge-base `82eccc83` = current tip)
**Verdict:** ✅ **Approve (carries + strengthened)** — the amendment adds a flat-pair cooldown safety guard; still fully fail-safe, parity/fallback unchanged.

Reviewed the `64a4deae → 3ecfa33a` delta with the amended-head method. It is a strict safety improvement to `_equity_hard_stop_try_reuse_replay_cache` plus a test — no change to the parity or fallback machinery I already verified.

**New guard — flat-pair cooldown correctness (another return-None-on-doubt path).** After loading the account manifest and held pairs, the gate now iterates `account_manifest["panic_flatten_events"]` and, for any marker whose symbol *is supported for coin replay* but is *not in the held-pair set*, rejects reuse (`return None` → authoritative full replay). Rationale (from the added comment): the replay treats every supported panic marker as a required replay pair, but this slice only loads matrices for currently-held pairs — a marker for a supported-but-now-flat coin would demand per-coin timeline values the synthesized rows can't provide, which would abort startup mid-replay or, worse, silently drop that coin's cooldown/no-restart evidence. Rejecting up front means the full replay reconstructs the flat-pair cooldown authoritatively. This directly serves the "reuse panic supervisor" / "held protective readiness" 9-Goals: a coin that panicked and is now flat still honors its cooldown, and reuse can never lose that evidence.

Note the guard is correctly scoped: a marker for an *unsupported* symbol is skipped (the full replay ignores it too, so reuse stays valid) — so this doesn't over-reject.

**Test pins it.** The new case persists a cache carrying a panic marker for a supported flat coin (`ETH/USDT:USDT`, not held) and asserts `_equity_hard_stop_try_reuse_replay_cache(...) is None`; then, with that symbol made unsupported, asserts reuse is allowed again (`is not None`) — proving both the rejection and the no-over-reject boundary.

**Verified by running the tests** at this head (main-clone Rust `.so`): the reuse suite passes — `test_hsl_cache_reuse_reaches_state_identical_to_full_replay` (the end-to-end parity: cache hit → identical coin HSL state) and `test_hsl_cache_reuse_falls_back_on_missing_or_incompatible_cache` (now also covering the flat-pair-marker rejection). I confirmed the guard is the only source change (the fresh-coverage-proof, strict `fill_coverage_proven`, `_try_load` validation + config_digest, watermark agreement, panic-in-extension-gap fallback, `_hsl_replay_extend_*`, current-position reconciliation, and `_hsl_replay_timeline_rows_from_cache` parity path are all unchanged from `64a4deae`).

Everything from my `64a4deae` review stands: reuse is a pure fail-safe accelerator — a hit is proven to yield the identical coin HSL state as the authoritative replay, and every unproven/mismatched/incomplete condition (now including a flat-pair panic marker) falls back to the full exchange-derived replay. Re-approving `3ecfa33a`.

PR #1127, opened 2026-07-07
Use max(raw, ema) for the HSL no-restart trigger in Rust and live

enarjord:
First behavior slice of the clarified episode/RED/no-restart contract (#1122, plan B2.1). The permanent no-restart halt now trips on `max(drawdown_raw, drawdown_ema) >= hsl_no_restart_drawdown_threshold` in **both** live and backtest; previously both evaluated raw drawdown only, under-triggering whenever sustained smoothed damage exceeded the instantaneous drawdown at the stop sample (e.g., deep drawdown with partial recovery right before the scope flattened). The RED/panic-now trigger is **unchanged**: recon showed the shared Rust runtime already implements the contract's `min(raw, ema)` tier score — this PR pins that with a regression test instead of touching it.

Centralization (per the guiding contract added in #1124): the predicate is owned by Rust and consumed by both paths —

- `ehsl::no_restart_triggered(policy, drawdown_raw, drawdown_ema, threshold)` in the shared `equity_hard_stop_loss` module, with the policy semantics unchanged (`always` → never latch, `never` → always latch, `threshold` → the max comparison).
- Backtest: `hsl_no_restart_latched` delegates to it; `HardStopStep` and the backtest stop snapshots now carry `drawdown_ema` (populated from the shared runtime state at snapshot time), and both evaluation sites (pside and coin paths) pass it. Rust unit tests updated, including the decisive `("threshold", raw=0.02, ema=0.11, thr=0.10) → latched` case.
- Live: exposed as `pbr.hsl_no_restart_triggered` via PyO3; `_equity_hard_stop_no_restart_latched` delegates, and all four finalize/latch call sites pass the matching `drawdown_ema` from their metrics/stop events (coin metrics finalize, stop-metrics finalize, recorded no-restart stop, coin stop event).

Tests/evidence:
- PyO3 predicate verified end-to-end after a fingerprint-stamped rebuild (all policies + invalid-policy ValueError).
- New contract pins in `tests/test_hsl_metric_regression.py`: max-trigger semantics across the three policies (including ema-trips-while-raw-recovered), and the RED `min(raw, ema)` pin — a raw flash-crash spike against a calm 60m EMA must not trigger RED (cites the #1122 rationale; users wanting raw sensitivity set `hsl_ema_span_minutes=1`).
[...]

PR #1122, opened 2026-07-06
Clarify HSL episode/RED/no-restart contract in the audit plan

enarjord:
Plan/docs only — records the clarified HSL episode/RED/no-restart contract in the fable-audit action plan. No behavior change; this is the target contract for the episode-anchoring and RED-supervision slices.

Contract points added (B2.1, with B2.7/B2.8/A2.1 and the decision recaps revised to match):

1. **Episode end** = the fill timestamp that flattens the full scoped HSL position set (unified: whole account; pside: that side; coin: that coin+pside) — explicitly not the RED breach, panic submit, panic create, or panic initiation timestamp.
2. **Split RED concepts**: `red_active_now` (the only authorizer of new panic orders) vs `red_seen_in_episode` (post-episode cooldown/restart evidence). RED seen earlier must not keep generating panic orders once the current sample is out of RED.
3. **Trigger formulas**: RED/panic-now uses `min(drawdown_raw, drawdown_ema) >= red_threshold` (EMA damps V-shaped flash crashes; raw sensitivity available via `hsl_ema_span_minutes=1`). No-restart uses `max(drawdown_raw, drawdown_ema) >= hsl_no_restart_drawdown_threshold` (permanent halt is conservative).
4. **Cooldown** starts at scoped episode end and activates when `red_seen_in_episode` was true.
5. **Partial flattening** keeps the episode and its drawdown tracker active; no panic resubmission on stale RED; leftover reduce-only orders follow normal flat-position cleanup.
6. **Incomplete history policy**: `always` may warn/degrade when pre-episode history is missing but the current episode is proven; `threshold`/`never` require full configured lookback coverage; missing current-episode evidence blocks startup unless a loud, per-run, documented dangerous CLI override is used.

Removed/revised older wording that implied: cooldown anchors at the RED timestamp (B2.7, user-decisions bullet, resolved-decisions bullet), and "permanently halt after any RED breach" (now "after any episode in which RED was active"). The remaining-details list now records what the contract resolves (equity ≤ 0 semantics, multiple crossings per episode, late coverage) and what stays open (flat-detection tolerance, override CLI surface).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1116, opened 2026-07-06
Repair silently broken risk/HSL regression tests

enarjord:
Repairs the risk/unstuck/HSL regression-test surfaces that broke silently as intentional behavior changes landed on v8. The repo's GitHub CI workflow is a no-op (`run: 'true'`), so none of these were caught at merge time; they were found by running the full suite locally while working the fable-audit action plan.

Baseline before this PR (fresh `origin/v8`, rebuilt Rust extension): **26 failing tests** across 6 files. After this PR: **3 known failures remain**, all out of scope (listed at the bottom).

Scope (mostly test-only; one small fail-loud guard and one doc parity fix):

- `tests/test_hsl_coin_mode.py`: `test_coin_hsl_check_tp_only_orange_skips_flat_symbols` still pinned the pre-#1098 contract. Rewritten as `..._blocks_flat_initial_entries` pinning the A2.2 contract: ORANGE `tp_only_with_active_entry_cancellation` also forces the mode for flat symbols in scope, blocking initial entries.
- `tests/test_unstucking_safeguards.py`:
  - DummyBot fixture predated the strict string bot params; `bot_value`/`bp` neutral `0.0` fallbacks now fail the strict normalizers. Added string defaults for `risk_twel_enforcer_policy`, `hsl_restart_after_red_policy`, `hsl_orange_tier_mode`, `hsl_panic_close_order_type`; removed a duplicated misindented `_bp_defaults` line (merge artifact).
  - Eight unified-path `initialize_from_history` tests inherited the template `hsl_signal_mode=coin` default (C1 alignment) and died mid-replay; they now pin `hsl_signal_mode="unified"` explicitly, matching sibling tests.
  - `..._preserves_no_restart_peak_across_replay_reset` expected pre-#1107 semantics where a panic marker alone reconstructed a RED stop. Re-authored with a shorter cooldown so the second episode has a pre-drop sample and reconstructs a *confirmed* RED per the #1107 contract; the test still proves its core point — the no-restart peak (110) survives episode resets.
- `src/passivbot_hsl.py`: `_equity_hard_stop_initialize_from_history` now fails loudly when invoked with `signal_mode=coin` (which must use the coin initializer) instead of surfacing a misleading mid-replay "unified signal mode requires unrealized_pnl_total" error. Production dispatch already routes coin mode to the coin initializer, so no live behavior change; new regression test.
[...]

PR #1128, opened 2026-07-07
Track red_active_now and red_seen_in_episode in the HSL runtime

enarjord:
Groundwork slice for the B2.1 red split (#1122 contract, item b). The shared Rust runtime now distinguishes the three RED concepts that were previously collapsed into the tier latch:

- **`HardStopStep.red_active_now`** — whether *this* sample's `min(raw, ema)` score crosses `red_threshold`, independent of latching. Per the clarified contract, this is the only signal that may authorize new panic orders; a latched tier alone must not.
- **`HardStopState.red_seen_in_episode`** — set once any sample in the current episode was red-active; cleared only on episode reset (both wrapper resets replace the state wholesale). Post-episode cooldown/restart evidence.
- **`red_latched`** — unchanged display/tier-pinning semantics.

Surface: PyO3 runtime `apply_sample` dict + `red_seen_in_episode()` accessor; the stateless `equity_hard_stop_step_py` gains an optional `red_seen_in_episode` kwarg (conservative default: `red_latched`, since a latched red was necessarily seen); the live `apply_sample`/`apply_coin_metrics_sample` metrics dicts carry both fields; and — applying the #1127 lesson proactively — the hermetic mock runtime and fake step fixtures mirror the new semantics exactly.

**Zero behavior change in this slice**: the tier latch still pins display red, and panic supervision still keys off the tier. The follow-up (behavior) slice rewires panic authorization to consume `red_active_now` only and moves the cooldown anchor to the scoped episode-end fill timestamp; landing the observability first keeps that PR small and lets these fields soak in events/tests.

Evidence:
- Rust unit tests: active-now goes false on recovery while seen persists and the latch pins the tier; replay-mode (no-latch) tracking; reset clears both.
- PyO3 verified end-to-end after a fingerprint-stamped rebuild.
- Python pins through the real runtime in `test_hsl_metric_regression.py` (latch test now asserts the split across crash → recovery).
- 218 passed across the HSL suites; full suite shows only the 3 known pre-existing out-of-scope failures (see #1116). `git diff --check` clean.


enarjord:
@vigilpbclaw-hash Hermes, Claude, Codex: requesting review of this PR (head 5cfcc38efa3ca5043118344e75fa5f055de56016).

Reviewer note: intended as a zero-behavior-change observability slice — the claims to verify are (1) no consumer's decision changes (tier/latch semantics byte-identical; the new fields are write-only additions to steps/metrics), and (2) the stub mirrors are exact (the #1127 finding class). The follow-up behavior slice will gate panic authorization on red_active_now and move the cooldown anchor to the episode-end fill, per the plan tracker note in this diff.

vigilpbclaw-hash:
PR: #1128
Head SHA: 5cfcc38efa3ca5043118344e75fa5f055de56016
Base SHA/branch: d9a4b1a8db15620a07cc07b6a11fa341c79a2513/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass: This PR is a groundwork/observability split for HSL RED semantics, separating `red_active_now`, `red_seen_in_episode`, and existing `red_latched` without changing panic/tier behavior yet. That is a good fit for the stated goal: the new fields are introduced at the Rust source of truth, surfaced through PyO3 and live metric dicts, and mirrored in the hermetic mock fixtures, while existing live panic authorization still keys off the old latch as documented for this slice.

Evidence reviewed:
- `passivbot-rust/src/equity_hard_stop_loss.rs`: `red_active_now` is computed from the current `drawdown_score` before latch pinning, while `red_seen_in_episode` only persists after a current RED sample and resets with fresh state.
- `passivbot-rust/src/python.rs` and `src/passivbot_hsl.py`: new fields are propagated in runtime/stateless outputs and live metric snapshots; existing `red_latched`/tier behavior is left intact.
- Regression tests cover latch-vs-current RED recovery and mock/runtime payload compatibility.

Validation run:
- `python3 -m py_compile src/passivbot_hsl.py tests/test_hsl_metric_regression.py tests/test_unstucking_safeguards.py` — passed.
- `maturin develop --release` in `passivbot-rust` — passed and installed the reviewed extension for local tests.
- `PYTHONPATH=src python3 -m pytest tests/test_hsl_metric_regression.py tests/test_unstucking_safeguards.py::test_hard_stop_apply_sample_delegates_to_rust tests/test_unstucking_safeguards.py::test_hard_stop_apply_sample_unified_uses_total_signal tests/test_unstucking_safeguards.py::test_hard_stop_apply_sample_same_minute_recomputes_when_inputs_change -q` — 13 passed.

Note: direct `cargo test ... --lib` still fails in this macOS review environment at the PyO3 test-link stage with unresolved Python symbols; I did not count that as a branch blocker because the project’s PyO3 build path via `maturin develop --release` succeeded and the targeted Python regression tests passed against the rebuilt reviewed extension.

enarjord (commented):
**PR:** #1128 — Track red_active_now and red_seen_in_episode in the HSL runtime
**Head SHA:** `5cfcc38efa3ca5043118344e75fa5f055de56016`
**Base:** v8 (merge-base `d9a4b1a8` = current tip — branch is right at HEAD)
**Scope:** `passivbot-rust/{equity_hard_stop_loss,backtest,python}.rs`, `src/passivbot_hsl.py`, `tests/{test_hsl_metric_regression,test_unstucking_safeguards}.py`, audit doc (+145/-3). MAX-care coin-HSL (contract c, #1122 item b) — but an **observability-only** slice.
**Verdict:** ✅ **Approve** — no findings. A genuinely zero-behavior-change RED-concept split: the new fields are computed and surfaced but consumed by no decision, and the tier ladder is byte-identical to before. Verified statically and by running the suites against a fresh fingerprint-matched build.

This is the groundwork slice of the B2.1 RED split: it distinguishes `red_active_now` (this sample's `min(raw,ema)` crosses `red_threshold`, no latch), `red_seen_in_episode` (sticky within the episode), and the unchanged `red_latched` (display/tier pin).

**Zero behavior change — proven, not assumed.**
- `red_active_now` is a **verbatim extraction** of the pre-existing RED tier condition: the code that read `drawdown_score + cmp_eps >= config.red_threshold` inline in the `next_tier` ladder is now `let red_active_now = drawdown_score + cmp_eps >= config.red_threshold;` reused in the identical ladder (`if red_latched → Red else if red_active_now → Red else if …orange…`). The tier is computed exactly as before.
- **No consumer.** `grep` across `src/` and `passivbot-rust/src/` shows the only reads of the new fields are the write into `state.red_seen_in_episode`, the tier-ladder reuse (identical), and pure surface (PyO3 `apply_sample` dict + `red_seen_in_episode()` accessor + the two live metrics dicts). Nothing keys a decision (panic authorization, cooldown, latch) off either field — panic supervision still keys off the tier, as the body states.

**Episode-scoping is correct (matters for the follow-up behavior slice).** `red_seen_in_episode` is set on any red-active sample and cleared on episode reset:
- Live: both PyO3 resets do `self.state = ehsl::HardStopState::default()` (wholesale → field false).
- Backtest: both restart paths (`try_restart_after_hard_stop` pside + coin) set `runtime.state = None`, and the next sample re-inits a fresh `HardStopState::default()` via `get_or_insert_with(...)` → field false. So the memory is correctly episode-scoped in both engines, not just the unit test.
- The flat/synthetic cached-step paths (backtest + the zero-drawdown early return) set `red_active_now: false`, which is correct (score 0 < threshold).

**Struct-completeness / KeyError safety.** Adding `red_active_now` to `HardStopStep` forces every constructor to set it — the compiler guarantees this, and the two extra sites (backtest flat path, main step) are handled. On the Python side, `_equity_hard_stop_apply_sample`/`_apply_coin_metrics_sample` now unconditionally read `step["red_active_now"]`/`step["red_seen_in_episode"]`; the real runtime dict always carries them, and the author updated the three monkeypatched `fake_apply_sample` fixtures + the hermetic mock runtime's `_cached_step` accordingly. I confirmed no other test feeds a step-dict missing the keys (the `drawdown_score` dicts in the live/report suites are metrics-output, not step-input, and downstream report/event code reads neither new field).

**The #1127 lesson applied proactively.** The hermetic `_make_mock_pbr()` runtime mirrors the new semantics exactly — `_state_reset`/`reset_state_keep_peak` clear `_red_seen_in_episode`, and the apply mirror computes `red_active_now = drawdown_score >= red_threshold`, sets `_red_seen_in_episode` when active, and keeps the tier logic `if red_latched or red_active_now` (identical structure to Rust). So the split stays hermetic/order-independent — no isolation regression this time.

**Value-safety.** Both new fields are booleans (labels), no magnitudes — safe for any downstream event/report surface.

**Verified by running it (fresh maturin build, fingerprint-matched):**
- Rust unit tests present and correct: `red_active_now_splits_from_seen_and_latch` (active goes false on recovery while seen persists and the latch pins the tier) and `red_seen_in_episode_tracks_without_latching` (replay/no-latch).
- `test_hsl_metric_regression.py` — the latch test now pins the split through the **real runtime**: crash → `red_active_now True`/`seen True`; full recovery → `red_active_now False` but `seen True` and tier still red (latched). Passes.
- Ran **436 tests green** against the fresh `.so` with no KeyError: `test_hsl_metric_regression` + `test_unstucking_safeguards` + `test_equity_hard_stop_rolling_peak` (129), the live/report suites carrying `drawdown_score` fixtures (`test_live_event_bus`/`test_live_performance_report`/`test_live_smoke_report`/`test_monitor_tui`/`test_hsl_startup_preview`, 212), and the full `test_hsl_coin_mode` contract-c suite (95).

### Gate note (for the follow-up behavior slice)
This slice is inert, so the load-bearing correctness of `red_seen_in_episode`'s episode-scoping (verified here in both engines) and of `red_active_now` as the *sole* panic-authorization signal only bites when the next PR rewires panic authorization to consume `red_active_now` and moves the cooldown anchor to the scoped episode-end fill timestamp. That PR is the MAX-care behavior change — it should re-pin: a latched-but-recovered tier must NOT authorize new panic orders (red_active_now False), and cooldown/no-restart evidence must key off `red_seen_in_episode`. Hermes should confirm the Rust unit tests + real-runtime split pin on CI.

Net: a clean, value-safe, zero-behavior-change observability split — `red_active_now` extracted identically from the existing RED condition, `red_seen_in_episode` correctly episode-scoped in live and backtest, consumed by nothing, with strong Rust + real-runtime pins and the hermetic mock kept faithful. Approving `5cfcc38e`.

enarjord:
Codex review: approve.

Scope reviewed: full PR at head 5cfcc38efa3ca5043118344e75fa5f055de56016 against v8.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

What I checked:
- red_active_now is computed from the current drawdown_score before latch pinning, and red_seen_in_episode only flips after a current RED sample, then resets with fresh runtime state.
- Existing behavior consumers still key off red_latched/tier; the new fields are surfaced through PyO3/live metric dicts and tests, but do not authorize panic/restart/cooldown behavior in this slice.
- The hermetic passivbot_rust test stubs mirror the new runtime methods/output keys, including cached-step payloads.

Validation run:
- git diff --check origin/v8...HEAD passed.
- cargo check in passivbot-rust passed.
- maturin develop --release installed the reviewed extension into the shared review venv.
- python -m pytest tests/test_hsl_metric_regression.py tests/test_unstucking_safeguards.py -q passed.
- python -m pytest tests/test_hsl_coin_mode.py -q passed.
- Direct PyO3 runtime probe confirmed recovery keeps red_seen_in_episode=True while red_active_now=False under a latched red tier.
- Real backtest smoke passed: passivbot backtest configs/ema_anchor_smoke_test.json.

Note: raw cargo test -q red_ still fails in this macOS review environment at the PyO3 test-binary link step with unresolved Python symbols. I do not consider that a PR blocker because cargo check, maturin develop --release, and the Python regression tests against the rebuilt extension all pass.

PR #1129, opened 2026-07-07
Authorize HSL panic orders from red_active_now only

enarjord:
Second behavior slice of the clarified episode/RED contract (#1122, plan B2.1), building on the #1128 red-split groundwork: **a latched RED episode no longer keeps emitting panic closes until flat** — panic emission is authorized only while the *current* drawdown sample is in RED, in both live and backtest.

**Live**:
- `_equity_hard_stop_coin_needs_panic_supervision` consumes the latest sample's `red_active_now` when latched; no sample yet against a latched state stays protective.
- Both RED supervisors (coin + unified/pside) refresh a metrics sample each cycle. A recovered sample downgrades the scope's forced modes to `tp_only_with_active_entry_cancellation` for the remainder of the episode — entries stay blocked (A2.2 semantics), normal closes proceed, no new panic orders — and panic re-engages if RED re-activates (contract point 5: partial flatten + recovery must not keep panicking). **Any refresh failure keeps panic modes with a warning**: recovery must be *proven* by a fresh sample, so degradation is always in the protective direction (this is also exercised by the orchestration test's minimal fixture).
- Flat-scope stop finalization, cooldown, and no-restart accounting are unchanged — they use the episode's RED evidence (`red_seen_in_episode` / latch), not the moment-to-moment signal.
- New pside-wide `_equity_hard_stop_set_red_paused_runtime_forced_modes`, bound on `Passivbot` and pinned in the AST + fake-bot binding tests.

**Backtest parity**: `HardStopPsideRuntime` records the latest step's `red_active_now` (from the shared runtime step introduced in #1128); the pside and coin mode overrides emit `TradingMode::Panic` only while it is true, falling back to `TpOnly` for the rest of a latched red episode.

**Deliberately unchanged / recorded**: the live stop anchor already uses the latest panic-fill timestamp (the flattening fill for bot-flattened episodes); the manual-flatten anchor refinement and the incomplete-history policy remain tracked as follow-ups. The tracker also records a **pre-existing** backtest ORANGE flat-symbol parity gap observed while implementing (`apply_orange_override`'s has-pos gate diverges from the live A2.2 contract) for its own fix — intentionally not bundled here.

Evidence:
- Predicate pins: no-metrics → protective; active → authorize; recovered → pause; halted repanic-reset unaffected. Paused-mode helper coverage (position/order/active symbols all forced tp-only).
- 251 passed across the affected suites; full suite shows only the 3 known pre-existing out-of-scope failures (see #1116); extension rebuilt fingerprint-stamped before runs.
- CHANGELOG entry (user-facing behavior change in a risk gate, both live and backtest).
- `git diff --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- HSL panic orders are now authorized only while the CURRENT drawdown sample
+  is in RED (`red_active_now`), in both live and backtest. Previously a
+  latched RED episode kept emitting panic closes until the scope was flat
+  even after the drawdown recovered; now a recovered sample pauses panic
+  emission for the remainder of the episode while entries stay blocked
+  (`tp_only_with_active_entry_cancellation`), and panic resumes if RED
+  re-activates. Flat-scope stop finalization, cooldown, and no-restart
+  accounting are unchanged and still use the episode's RED evidence.
+
 - The HSL no-restart (permanent halt) trigger now evaluates
   `max(drawdown_raw, drawdown_ema)` against
   `hsl_no_restart_drawdown_threshold` in both live and backtest, instead of
--- a/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
+++ b/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
@@ -898,6 +898,20 @@ Remaining implementation details:
       this slice: the tier latch still pins display red; the follow-up slice
       rewires panic authorization to consume `red_active_now` only and moves
       the cooldown anchor to the scoped episode-end fill timestamp.
+      Implemented (panic authorization): live coin and unified/pside RED
+      supervisors now refresh a sample each cycle and emit panic modes only
+      while `red_active_now`; a recovered sample downgrades the scope to
+      `tp_only_with_active_entry_cancellation` for the remainder of the
+      episode and re-engages panic if RED re-activates. Backtest parity: the
+      pside/coin mode overrides gate `TradingMode::Panic` on the runtime's
+      latest `red_active_now` and fall back to tp-only. Live stop anchors
+      already use the latest panic-fill timestamp (the flattening fill for
+      bot-flattened episodes); the manual-flatten anchor refinement and the
+      incomplete-history policy remain open.
+      Observed while implementing (pre-existing, not this slice): the
+      backtest ORANGE `TpOnly` override still skips flat symbols
+      (`apply_orange_override` has-pos gate), which diverges from the
+      live A2.2 contract; needs its own parity fix.
 - [x] Dynamic WEL and snapped/raw balance docs/tests.
       Make `reduce_overweight` use dynamic currently-tradable slot count, keep
       snapped/raw balance separation, and add high-hysteresis warning/preflight
--- a/passivbot-rust/src/backtest.rs
+++ b/passivbot-rust/src/backtest.rs
@@ -460,6 +460,10 @@ fn hsl_no_restart_latched(
 struct HardStopPsideRuntime {
     state: Option<ehsl::HardStopState>,
     tier: ehsl::HardStopTier,
+    // B2.1 red split: whether the LATEST sample crossed RED. Only this may
+    // authorize panic-mode order emission; a latched red tier alone keeps
+    // entry blocking but no new panic orders.
+    red_active_now: bool,
     rolling_peak_strategy_pnl: VecDeque<(u64, f64)>,
     halted: bool,
     no_restart_latched: bool,
@@ -3248,7 +3252,15 @@ impl<'a> Backtest<'a> {
         }
         match runtime.tier {
             ehsl::HardStopTier::Red => {
-                *mode = Some(orchestrator::TradingMode::Panic);
+                if runtime.red_active_now {
+                    *mode = Some(orchestrator::TradingMode::Panic);
+                } else {
+                    // B2.1 red split: RED seen this episode but the current
+                    // sample recovered. Block entries for the remainder of
+                    // the episode without authorizing new panic orders,
+                    // matching the live supervisor's paused-red mode.
+                    *mode = Some(orchestrator::TradingMode::TpOnly);
+                }
             }
             ehsl::HardStopTier::Orange => {
                 let target = self.hard_stop_orange_target_mode_pside(pside);
@@ -3280,7 +3292,15 @@ impl<'a> Backtest<'a> {
         }
         match runtime.tier {
             ehsl::HardStopTier::Red => {
-                *mode = Some(orchestrator::TradingMode::Panic);
+                if runtime.red_active_now {
+                    *mode = Some(orchestrator::TradingMode::Panic);
+                } else {
+                    // B2.1 red split: RED seen this episode but the current
+                    // sample recovered. Block entries for the remainder of
+                    // the episode without authorizing new panic orders,
+                    // matching the live supervisor's paused-red mode.
+                    *mode = Some(orchestrator::TradingMode::TpOnly);
+                }
             }
             ehsl::HardStopTier::Orange => {
                 let target = self.hard_stop_orange_target_mode_pside(pside);
@@ -3568,6 +3588,7 @@ impl<'a> Backtest<'a> {
         let runtime = &mut self.hard_stop_pside[pside];
         let prev_tier = runtime.tier;
         runtime.tier = step.tier;
+        runtime.red_active_now = step.red_active_now;
 
         if step.tier == ehsl::HardStopTier::Red {
             if prev_tier != ehsl::HardStopTier::Red {
@@ -3765,6 +3786,7 @@ impl<'a> Backtest<'a> {
         let runtime = &mut self.hard_stop_coin[pside][idx];
         let prev_tier = runtime.tier;
         runtime.tier = step.tier;
+        runtime.red_active_now = step.red_active_now;
 
         if step.tier == ehsl::HardStopTier::Red {
             if prev_tier != ehsl::HardStopTier::Red {
--- a/src/passivbot.py
+++ b/src/passivbot.py
@@ -1855,6 +1855,9 @@ class Passivbot:
     _equity_hard_stop_set_red_runtime_forced_modes = (
         pb_hsl._equity_hard_stop_set_red_runtime_forced_modes
     )
+    _equity_hard_stop_set_red_paused_runtime_forced_modes = (
+        pb_hsl._equity_hard_stop_set_red_paused_runtime_forced_modes
+    )
     _equity_hard_stop_set_coin_runtime_forced_mode = (
         pb_hsl._equity_hard_stop_set_coin_runtime_forced_mode
     )
--- a/src/passivbot_hsl.py
+++ b/src/passivbot_hsl.py
@@ -5505,7 +5505,17 @@ def _equity_hard_stop_coin_needs_panic_supervision(
     if not self._equity_hard_stop_enabled(pside):
         return False
     if state["runtime"].red_latched() and not state["halted"]:
-        return True
+        # B2.1 contract (clarified 2026-07-06): only the CURRENT sample being
+        # in RED may authorize new panic orders. A latched tier with a
+        # recovered sample keeps the episode's entry blocking (forced-mode
+        # downgrade handled by the supervisor) but must not keep submitting
+        # panic orders.
+        last_metrics = state.get("last_metrics")
+        if last_metrics is None:
+            # No sample yet against the latched state: stay protective until
+            # the next sample proves recovery.
+            return True
+        return bool(last_metrics.get("red_active_now", True))
     return bool(state["halted"] and state["cooldown_repanic_reset_pending"])
 
 
@@ -5534,6 +5544,31 @@ def _equity_hard_stop_set_red_runtime_forced_modes(self, pside: str) -> None:
             previous_modes=previous,
             modes=forced,
             reason_code="hsl_red_runtime_forced_modes",
+        )
+
+
+def _equity_hard_stop_set_red_paused_runtime_forced_modes(self, pside: str) -> None:
+    """Episode entry blocking without panic emission (B2.1 red split).
+
+    Used while a red-seen episode is still open but the current sample is no
+    longer in RED: entries stay cancelled/blocked, normal closes may proceed,
+    and no new panic orders are authorized.
+    """
+    previous = dict(getattr(self, "_runtime_forced_modes", {}).get(pside, {}) or {})
+    forced = {}
+    symbols = set(self.positions.keys()) | set(self.open_orders.keys()) | set(self.active_symbols)
+    for symbol in symbols:
+        forced[symbol] = "tp_only_with_active_entry_cancellation"
+    self._runtime_forced_modes[pside] = forced
+    if previous != forced:
+        _emit_runtime_forced_mode_changed_event(
+            self,
+            pside=pside,
+            action="replace",
+            symbols=forced.keys(),
+            previous_modes=previous,
+            modes=forced,
+            reason_code="hsl_red_paused_runtime_forced_modes",
         )
 
 
@@ -6029,7 +6064,50 @@ async def _equity_hard_stop_run_red_supervisor(self) -> None:
             if not active_red_psides:
                 return
             for pside in active_red_psides:
-                self._equity_hard_stop_set_red_runtime_forced_modes(pside)
+                # B2.1 contract: refresh the sample so recovery is observable
+                # mid-supervision; only red_active_now authorizes continued
+                # panic emission for the episode. Any refresh failure keeps
+                # panic modes: recovery must be PROVEN by a fresh sample.
+                metrics = None
+                try:
+                    signal_mode = self._equity_hard_stop_signal_mode()
+                    upnl_total = (
+                        float(await self._calc_upnl_sum_strict())
+                        if signal_mode == "unified"
+                        else None
+                    )
+                    metrics = self._equity_hard_stop_apply_sample(
+                        pside,
+                        int(self.get_exchange_time()),
+                        float(self.get_raw_balance()),
+                        float(self._equity_hard_stop_realized_pnl_now()),
+                        float(self._equity_hard_stop_realized_pnl_now(pside)),
+                        float(await self._calc_upnl_sum_strict(pside)),
+                        unrealized_pnl_total=upnl_total,
+                    )
+                except asyncio.CancelledError:
+                    raise
+                except Exception as exc:
+                    logging.warning(
+                        "[risk] HSL[%s] RED supervisor sample refresh failed; "
+                        "keeping panic modes (recovery must be proven) | "
+                        "error=%s: %s",
+                        pside,
+                        type(exc).__name__,
+                        exc,
+                    )
+                if metrics is not None and not bool(
+                    metrics.get("red_active_now", True)
+                ):
+                    logging.info(
+                        "[risk] HSL[%s] RED no longer active on current sample; "
+                        "pausing panic emission for the remainder of the episode "
+                        "(entries stay blocked)",
+                        pside,
+                    )
+                    self._equity_hard_stop_set_red_paused_runtime_forced_modes(pside)
+                else:
+                    self._equity_hard_stop_set_red_runtime_forced_modes(pside)
             self._equity_hard_stop_refresh_halted_runtime_forced_modes()
             try:
                 to_cancel, to_create = (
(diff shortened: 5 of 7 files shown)


enarjord:
@vigilpbclaw-hash Hermes, Claude, Codex: requesting review of this PR (head 50b0c7ab9d57ef8a7a76844503b3e4633c502b58).

Reviewer notes — this changes panic emission in a MAX-care risk gate, so scrutinize the direction of every degradation: (1) the change only ever REDUCES panic emission relative to before, and only when a fresh sample proves min(raw, ema) is back under red_threshold; every uncertainty path (no sample, refresh failure) keeps the previous panic behavior; (2) entries remain blocked throughout the episode via the tp-only downgrade, so the recovered-sample state is strictly between old-panic and normal trading; (3) backtest and live must degrade identically — the backtest gate consumes the same shared runtime step field; (4) stop finalization/cooldown/no-restart are intentionally untouched (episode evidence, not moment signal). If you can construct a scenario where the scope ends an episode WITHOUT finalization because panic paused (e.g., supervisor exits while non-flat and nothing re-checks), that's the finding that matters most.

vigilpbclaw-hash:
PR: #1129
Head SHA: 50b0c7ab9d57ef8a7a76844503b3e4633c502b58
Base SHA/branch: 65b73b776a69d5b7b6e0cc1f98ff97fcf678417e/v8
Scope reviewed: full PR
Verdict: needs fix

Findings:
- Important: `src/passivbot_hsl.py:6111` overwrites the recovered-sample pause back to panic for unified/pside RED episodes. In `_equity_hard_stop_run_red_supervisor`, the branch at lines 6099-6109 correctly calls `_equity_hard_stop_set_red_paused_runtime_forced_modes(pside)` when the fresh sample has `red_active_now == False`, but the very next statement calls `_equity_hard_stop_refresh_halted_runtime_forced_modes()`. That helper still treats any non-halted `runtime_red_latched(pside)` as full RED and unconditionally calls `_equity_hard_stop_set_red_runtime_forced_modes(pside)` (`src/passivbot_hsl.py:3927-3929`), replacing the paused `tp_only_with_active_entry_cancellation` modes with `panic`. As a result, the PR's main live pside/unified contract (latched RED + recovered current sample must stop authorizing new panic orders) does not hold; `calc_protective_panic_orders_to_cancel_and_create()` then runs with panic modes anyway. A concrete fix is to make the refresh helper preserve/derive the paused-red state from the current sample, or skip that refresh here except for truly halted sides, and add a regression test that exercises the full supervisor path (paused helper followed by the refresh) rather than only the helper in isolation.

Minor:
- The added Python tests cover coin-mode `_equity_hard_stop_coin_needs_panic_supervision` and the paused helper directly, but not the pside/unified supervisor sequence where the regression above occurs. Please add coverage for a recovered pside/unified RED sample proving the final runtime forced modes remain `tp_only_with_active_entry_cancellation` through the order-planning call.

Intent/architecture pass:
- The intended split is sound: keep the RED episode latch for entry blocking/finalization while using only the current sample to authorize new panic emission, with uncertainty degrading protectively. The implementation is close, but the existing centralized forced-mode refresher is still built around the old "latched RED means panic" abstraction; the new paused-red state should be represented in that abstraction or the supervisor should avoid invoking it in a way that erases the pause.

Validation run:
- `git diff --check origin/v8...HEAD` — clean.
- Merge dry-run into current `origin/v8` — clean.
- `python3 -m py_compile src/passivbot_hsl.py src/passivbot.py` — passed.
- Tried targeted pytest for the new tests, but this checkout skips collection because `passivbot_rust` extension is not available in the cron environment; I did not treat that environment limitation as a branch failure.

enarjord:
Codex review: changes requested.

Findings:
- Blocker: coin HSL can skip RED stop finalization after panic pauses on a recovered sample. In src/passivbot.py:5323-5330, the main loop only enters the coin RED supervisor when _equity_hard_stop_coin_red_active() is true. This PR changes that predicate through src/passivbot_hsl.py:5502-5519 to return false for a red-latched, non-halted coin once last_metrics.red_active_now is false. But flat confirmation/finalization for coin RED episodes is only performed inside _equity_hard_stop_run_coin_red_supervisor(), specifically src/passivbot_hsl.py:6138-6184. The regular _equity_hard_stop_check_coin() path at src/passivbot_hsl.py:5393-5488 updates metrics and clears only non-red tiers; it does not finalize a flat red-latched episode. Scenario: coin enters RED and latches, supervisor refreshes a recovered sample and sets tp_only_with_active_entry_cancellation at src/passivbot_hsl.py:6214-6226, then recomputes active via the same panic predicate and returns because red_active_now is false. If the remaining position later closes normally under tp-only, the next main loop calls _equity_hard_stop_check_coin(), still sees a latched red tier, but _equity_hard_stop_coin_red_active() stays false, so the supervisor is not re-entered and the episode never reaches the flat-confirmation/finalize path. That means cooldown/no-restart/latch finalization can be missed for exactly the recovered-before-flat case this PR introduces.

Suggested direction: split the coin predicates. Panic order emission should use red_active_now, but supervisor/finalization liveness should continue while the coin scope is red-latched and not halted, or the normal check path must perform equivalent flat confirmation/finalization for red-latched recovered episodes. Please add a regression covering: coin RED latches, red_active_now recovers while position remains open, forced mode downgrades to tp-only, position later becomes flat, and the episode is finalized/cooldown state is set without requiring RED to become active again.

Validation so far:
- git diff --check origin/v8...HEAD passed.
- Static review of the live coin supervisor/main-loop path above. I stopped before approval-level runtime validation because this blocker affects the core behavior contract.

enarjord (commented):
**PR:** #1129 — Authorize HSL panic orders from red_active_now only
**Head SHA:** `50b0c7ab9d57ef8a7a76844503b3e4633c502b58`
**Base:** v8 (merge-base `65b73b77` = current tip — branch is right at HEAD)
**Scope:** `passivbot-rust/src/backtest.rs`, `src/passivbot.py`, `src/passivbot_hsl.py`, `tests/{test_hsl_coin_mode,test_passivbot_hsl_bindings}.py`, CHANGELOG, audit doc (+235/-5). **The signal-bearing B2.1 behavior slice** (contract c, #1122) building on #1128 — MAX-care.
**Verdict:** ✅ **Approve** — no findings. A correct, fail-safe, parity-faithful implementation: a latched-but-recovered RED episode no longer emits new panic orders, recovery must be *proven* by a fresh sample, cooldown/no-restart still key off episode evidence, and re-activation correctly re-engages panic. Verified statically and by running the affected suites against a fresh fingerprint-matched build.

**What changes (exactly the contract I was tracking).** Panic emission is authorized only while the *current* sample is RED (`red_active_now`, from #1128), in both live and backtest. A latched tier whose current sample recovered keeps the episode's entry blocking but stops submitting panic orders; panic re-engages if RED re-activates.

**Live — the gate is fail-safe in the protective direction (verified).**
- `_equity_hard_stop_coin_needs_panic_supervision`: when latched & not halted, returns `last_metrics.get("red_active_now", True)` — **defaults True (protective)** when the key is absent, and returns True when `last_metrics is None` (no sample yet). Only a fresh sample that *proves* recovery (`red_active_now == False`) drops panic authorization.
- Both supervisors (coin + unified/pside) refresh a sample each cycle. On a proven-recovered sample → paused forced modes (`tp_only_with_active_entry_cancellation`: entries cancelled/blocked, normal closes proceed, no panic). On **refresh failure** → `metrics` stays `None` → the `if metrics is not None and not red_active_now` is False → **else branch keeps panic modes** (logged warning "recovery must be proven"). So every degradation is protective. Confirmed by inspection at both supervisor sites.
- `_equity_hard_stop_set_red_paused_runtime_forced_modes` forces `tp_only_with_active_entry_cancellation` across `positions ∪ open_orders ∪ active_symbols` and emits the value-safe `hsl_red_paused_runtime_forced_modes` event (labels only).
- **Cooldown / no-restart / flat-scope finalization unchanged** — the diffstat touches only the two supervisor regions, the needs-panic gate, and the new helper; finalize/cooldown/no-restart still consume episode evidence (`red_seen_in_episode` / latch), not the moment signal, exactly as the body claims.

**Re-activation is correctly handled (the subtle point I checked hardest).** In coin mode the supervisor's entry gate `_equity_hard_stop_coin_red_active()` keys off `needs_panic_supervision` (moment-scoped), so after recovery the supervisor exits — but the main loop calls `_equity_hard_stop_check()` → `_equity_hard_stop_check_coin()` **every cycle**, which loops over all active coins and calls `_equity_hard_stop_apply_coin_sample` → refreshes `last_metrics` with a fresh `red_active_now` independent of the supervisor. So when RED re-activates, `needs_panic_supervision` flips True next cycle → the supervisor re-runs → panic re-engages (contract point 5). During the paused window the `tp_only` forced mode is sticky (entries stay blocked), and `check_coin` will not re-assert panic on an already-latched coin (`metrics["tier"]=="red" and not prev_latched` guard skips it). The unified/pside path is even more conservative — its gate is `red_latched and not halted`, so the supervisor stays resident for the whole latched episode, toggling panic↔paused per sample. Both paths satisfy the contract.

**Backtest parity (verified statically; compiled clean).** `HardStopPsideRuntime` gains `red_active_now: bool` (struct derives `Default` → false; set alongside `tier` on every apply at both apply sites, so it is always fresh before any override reads it — same freshness guarantee `tier` already had; restart sets tier=Green making a stale value inert). Both mode-override sites emit `TradingMode::Panic` only if `red_active_now`, else `TradingMode::TpOnly` — the backtest analog of the live paused mode (no backtest notion of cancelling resting entry orders, so `TpOnly` ≙ `tp_only_with_active_entry_cancellation` for "entries off, closes on, no panic"). ORANGE is untouched.

**Verified by running it (fresh maturin build, fingerprint-matched):**
- 221 tests green across `test_hsl_coin_mode` + `test_passivbot_hsl_bindings` + `test_unstucking_safeguards` + `test_hsl_metric_regression`, no failures.
- The two decisive new pins pass individually: `test_coin_panic_supervision_requires_red_active_now` (no-sample → protective True; active → True; recovered → **False**; halted repanic-reset → True unaffected) and `test_red_paused_forced_modes_block_entries_without_panic` (tp-only forced across position/order/active symbols). Bindings pinned in both the AST test and the fake-bot binding.

**Value-safety.** New forced-mode events carry reason_code + symbols/modes (labels); supervisor logs print pside/symbol only. No magnitudes on any shareable surface.

### Env / gate notes for the two-reviewer gate
- `test_order_orchestration.py` and `test_passivbot_balance_split.py` are collection-gated in my env by ccxt 4.5.22 vs required 4.5.48 (pre-existing, #1116) — so the supervisor orchestration test (incl. the refresh-failure→keep-panic fixture) runs on CI, not locally. The refresh-failure path is correct by inspection, but Hermes should confirm that orchestration test on a correct-ccxt env.
- Ideal soak: a latched episode that recovers → verify panic stops while entries stay blocked (`tp_only`), then re-activates → verify panic re-engages, then flattens → verify finalize/cooldown fire off `red_seen_in_episode` (not the moment signal). And a backtest run showing the same latched→recover→pause→re-activate sequence for parity.
- The body records a **pre-existing** backtest ORANGE flat-symbol parity gap (`apply_orange_override`'s has-pos gate vs live A2.2) tracked for its own fix — I confirmed this PR does not touch ORANGE, so it's correctly deferred, not introduced here.

Net: a correct, fail-safe, parity-faithful panic-authorization split — panic keys off `red_active_now` only, recovery must be proven (any refresh failure keeps panic), the paused mode blocks entries without panic, re-activation re-engages via the persistent main-loop sampler, and cooldown/no-restart remain episode-scoped. Approving `50b0c7ab`.

enarjord:
Pushed ea8e860f addressing both findings:

**Codex blocker (finalization liveness)**: the regular `_equity_hard_stop_check_coin` now owns paused episodes — when the runtime is red-latched, not halted, and the current sample recovered, it keeps the tp-only forced mode and performs the same flat-confirmation + `finalize_coin_red_stop` sequence the supervisor would. Cooldown/no-restart accounting can therefore never be skipped because RED recovered before flat, and the supervisor remains panic-only (no long-lived supervision loop blocking the main loop during recovered episodes). Regression exactly per your spec: latch → recover with open position (panic pauses, `needs_panic_supervision` false) → normal flatten → two check cycles finalize with halt + cooldown set, RED never re-activates.

**Hermes important (refresher overwrite)**: `_equity_hard_stop_refresh_halted_runtime_forced_modes` now derives panic vs paused from the latest sample's `red_active_now` for latched non-halted psides (missing sample stays protective panic), so the paused tp-only modes survive the refresh that runs right after the supervisor sets them — and every other caller of the centralized refresher inherits the same semantics. Regression: paused modes survive the refresher; active/unknown samples still force panic.

Validation: 253 passed across the affected suites (`test_hsl_coin_mode`, `test_unstucking_safeguards`, `test_order_orchestration`, `test_hsl_metric_regression`, bindings); `git diff --check` clean.

@vigilpbclaw-hash Hermes, Claude, Codex: requesting delta re-review of 50b0c7ab..ea8e860f.

vigilpbclaw-hash:
PR: #1129
Head SHA: ea8e860f0e8472f99f0d817a3c993483a8287d78
Base SHA/branch: 65b73b776a69d5b7b6e0cc1f98ff97fcf678417e/v8
Scope reviewed: delta 50b0c7ab9d57ef8a7a76844503b3e4633c502b58..ea8e860f0e8472f99f0d817a3c993483a8287d78
Verdict: approve

Supersedes my prior review of head `50b0c7ab9d57ef8a7a76844503b3e4633c502b58`; this comment is SHA-specific to the current head above.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass:
- The delta addresses the intended B2.1 RED split in the right abstraction points: the centralized pside forced-mode refresher now derives panic vs paused from the latest `red_active_now` sample, and the coin-mode regular check path owns latched-but-recovered episodes so finalization/cooldown accounting cannot depend on the panic supervisor staying active. That preserves the intended split between episode evidence (latch/finalization/no-restart) and momentary panic authorization (`red_active_now`) without introducing a new exchange-action gate.

Evidence reviewed:
- `src/passivbot_hsl.py:3927` now prevents `_equity_hard_stop_refresh_halted_runtime_forced_modes()` from overwriting a recovered latched pside back to panic; missing/unknown sample remains protective panic.
- `src/passivbot_hsl.py:5507` keeps coin entries blocked under `tp_only_with_active_entry_cancellation` and performs flat confirmations/finalization from `_equity_hard_stop_check_coin()` when a red-latched coin has recovered before flattening.
- Added regressions cover both previously reported paths: recovered coin RED finalizes from the check path, and recovered pside RED paused modes survive the centralized refresher.

Validation run:
- `git diff --check origin/v8...origin/pr-1129` — clean.
- Merge dry-run into current `origin/v8` — clean.
- `python3 -m py_compile src/passivbot_hsl.py src/passivbot.py tests/test_hsl_coin_mode.py` — passed.
- Targeted pytest for the new/related tests could not collect in this cron checkout because `ccxt` is not installed in the environment (`ModuleNotFoundError: No module named 'ccxt'`); I did not treat that environment limitation as a branch finding.

enarjord (commented):
**Reviewer:** Claude (automated continuous reviewer) — SHA update
**PR:** #1129 — Authorize HSL panic orders from red_active_now only
**Head SHA:** `ea8e860f0e8472f99f0d817a3c993483a8287d78` (amended from previously-reviewed `50b0c7ab`)
**Base:** v8 (merge-base `65b73b77` = current tip)
**Verdict:** ✅ **Approve** — the amendment fixes two real correctness bugs in the recovered-episode path (a Codex blocker + a Hermes finding) that **my `50b0c7ab` review missed**. Both fixes are correct, mutually consistent, and verified on a fresh build.

Reviewed the `50b0c7ab → ea8e860f` delta (+176 test-heavy, `src/passivbot_hsl.py` +61 / `tests/test_hsl_coin_mode.py` +116). It closes two gaps I did not catch when I approved `50b0c7ab` with "no findings" — I'm calling that out explicitly.

**Bug 1 (Hermes) — the pside pause was clobbered back to panic. My miss.** At `50b0c7ab`, the unified/pside supervisor set paused (`tp_only`) modes, then immediately called `_equity_hard_stop_refresh_halted_runtime_forced_modes()`, which unconditionally re-applied `_equity_hard_stop_set_red_runtime_forced_modes(pside)` (= panic) for any latched non-halted pside — so the pside pause never actually held. My prior review claimed the pside path "toggles panic↔paused per sample"; it did not, because the trailing refresher overwrote the pause. **Fix:** the refresher now derives panic-vs-paused from `last_metrics.get("red_active_now", True)` — paused only on a proven-recovered sample, panic otherwise (protective default when no sample). Correct and protective. Pinned by `test_forced_mode_refresher_preserves_paused_red` (no sample → panic; active → panic; recovered → paused survives the refresher).

**Bug 2 (Codex, blocker) — a recover-before-flat episode skipped finalization/cooldown. My miss.** At `50b0c7ab`, the flat-confirmation → `finalize_coin_red_stop` (halt + cooldown + no-restart accounting) lived **only** in the coin supervisor. But once the current sample recovers, `needs_panic_supervision` goes False and the supervisor stops running. So a latched episode that recovered while the position was still open, then flattened normally under tp-only, would **never finalize** — cooldown/no-restart evidence silently lost. My prior review asserted "cooldown/no-restart/finalization unchanged" and treated it as correct; I failed to notice finalization was reachable only through the now-disengaged supervisor. **Fix:** `_equity_hard_stop_check_coin` now, when `red_latched and not halted and not red_active_now`, keeps entries blocked (`tp_only`) **and** performs the same flat-confirmation/finalization the supervisor would (`red_flat_confirmations`, `compute_coin_stop_event`, `finalize_coin_red_stop(finalized_without_order=True)` at ≥2 confirmations). So cooldown/no-restart can never be skipped just because RED recovered before flat.

**The two finalization paths are mutually exclusive — no double-finalize (verified).** The check-path block is guarded by `not bool(metrics.get("red_active_now", True))` (runs only when recovered; protective default keeps it off when the key is missing), while the supervisor only runs when `red_active_now`/no-sample (needs-panic True). Since `_equity_hard_stop_check()` runs before the supervisor each main-loop cycle: red-active → supervisor finalizes, check-path block skipped; recovered → check-path finalizes, supervisor doesn't run. No overlap. The block also mirrors the supervisor's has-position/blocking-order gating exactly and refreshes `state` after finalize.

**Verified by running it (fresh maturin build, fingerprint-matched):**
- The two new regression tests pass: `test_recovered_red_episode_finalizes_from_check_path` (crash→RED→panic; recover-with-open-position→tp_only + no supervision needed; flatten→two check cycles→**halt + cooldown**, RED never re-activates and the supervisor never runs) and `test_forced_mode_refresher_preserves_paused_red`.
- Full affected suites green: `test_hsl_coin_mode` + `test_passivbot_hsl_bindings` + `test_unstucking_safeguards` + `test_hsl_metric_regression` — **223 passed** (the 221 from `50b0c7ab` plus these 2). Bindings updated for the newly-consumed methods (`_equity_hard_stop_count_blocking_open_orders_symbol`, `_refresh_halted_runtime_forced_modes`, etc.).

Everything else from my `50b0c7ab` review still stands and is unchanged: panic authorized only while the current sample is `red_active_now`; recovery must be proven (any supervisor sample-refresh failure keeps panic); backtest parity (`Panic` only if `red_active_now` else `TpOnly`); value-safe labels. The amendment strictly *adds* the two missing protective behaviors — it doesn't alter the authorization semantics I verified.

### Gate note
The recovered-then-flat finalization and the pside-pause preservation are now the load-bearing correctness of this slice. Hermes should confirm on CI (ccxt-gated locally): the full crash→recover→flat→halt/cooldown sequence and that the pside refresher preserves the pause; ideally a backtest/live soak of a partial-flatten-then-recovery episode confirming cooldown/no-restart still fire and entries stay blocked without panic.

Net: the amendment fixes a real blocker (recovered-before-flat episodes now finalize via the check path) and a real pause-preservation bug (the refresher no longer clobbers paused modes) — both of which I missed at `50b0c7ab`. Fixes verified correct and mutually exclusive, full suite green. Re-approving `ea8e860f`.

enarjord:
Codex delta review: approve.

Scope reviewed: delta 50b0c7ab9d57ef8a7a76844503b3e4633c502b58..ea8e860f0e8472f99f0d817a3c993483a8287d78, plus the full PR behavior contract where the delta interacts with the prior panic-active-now change.

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

The prior Codex blocker is resolved. Coin mode no longer relies on the panic supervisor for recovered red-latched episode finalization: the regular coin HSL check path keeps tp-only entry blocking, performs flat confirmations, and finalizes halt/cooldown once the recovered episode is flat. That covers the red-latches, recovers while still open, later flattens normally case without needing RED to reactivate.

Hermes prior finding is also resolved. The centralized pside forced-mode refresher now uses latest last_metrics.red_active_now for latched non-halted psides: missing or active sample stays protective panic, while a proven recovered sample preserves the paused tp-only mode instead of clobbering it back to panic.

I also checked the two paths are separated in the intended direction: red-active or unknown still goes through panic supervision; proven-recovered coin episodes are owned by the regular check path for finalization; pside/unified supervisor remains resident and the refresher no longer overwrites the pause.

Validation run on ea8e860f:
- git diff --check origin/v8...HEAD passed.
- git diff --check 50b0c7ab..HEAD passed.
- cargo check in passivbot-rust passed.
- maturin develop --release installed the reviewed extension into the shared review venv.
- python -m pytest tests/test_hsl_coin_mode.py -q passed.
- python -m pytest tests/test_unstucking_safeguards.py tests/test_order_orchestration.py tests/test_hsl_metric_regression.py tests/test_passivbot_hsl_bindings.py -q passed.
- Real backtest smoke passed: passivbot backtest configs/ema_anchor_smoke_test.json.

Formal GitHub approval is not available from this token because it is treated as PR owner, so this comment is the Codex green review for head ea8e860f0e8472f99f0d817a3c993483a8287d78.

PR #1137, opened 2026-07-08
HSL: pure pside/unified timeline synthesis with parity trust boundary

enarjord:
## Summary

Third slice of the pside/unified replay-cache arc (#1135 write-only persistence → #1136 schema v5 per-pside pnl → this). Adds the pure, **unwired** synthesis helper `_hsl_replay_pside_timeline_rows_from_cache` plus the parity trust boundary the future reuse gate must build on — mirroring how the coin arc landed its synthesis slice.

## Changes

- `src/passivbot_hsl.py`: `_hsl_replay_pside_timeline_rows_from_cache(pair_arrays, account_arrays, *, current_balance)` synthesizes the exact aggregate row contract `_equity_hard_stop_initialize_from_history` consumes (`timestamp`, `balance`, `realized_pnl`, `realized_pnl_long/short`, `unrealized_pnl_long/short`, `is_flat`, `is_flat_long/short`):
  - balance anchored so the final minute equals the current balance (same convention as the coin synthesis);
  - per-pside realized cumsums from the schema-v5 `pnl_long`/`pnl_short` arrays;
  - per-pside upnl sums and flatness from cached pair matrices — psize is **signed** (shorts negative), so flatness uses `abs(psize) > 1e-12` (the parity test caught this during development);
  - the same fail-loud span/continuity/grid-alignment validations as the coin synthesis.
- **Documented critical precondition** (docstring + tracker): the future reuse gate must prove from fills that cached pairs were the ONLY pairs with in-window positions before trusting the upnl/flatness aggregates — mirroring coin-mode flat-pair rejection — else force full replay. Nothing is wired in this slice.

## Tests

- `test_pside_timeline_synthesis_matches_authoritative_rows`: field-for-field equality against the real `get_balance_equity_history` timeline (unified mode; one long + one short pair; realized events on both sides; flatness transitions on both psides; exact contract key-set pin).
- `test_pside_initializer_state_parity_on_contract_shaped_rows`: rows carrying only the contract fields drive the pside/unified initializer to a state identical to authoritative-shaped rows (extra fields present), through a mid-replay RED crossing with pinned final sample.

Full local suite green except the 3 known out-of-scope failures.

## Noted follow-up (tracked)

While building the state-parity fixture: the pside/unified replay still derives cooldown/no-restart only from panic markers — a RED-seen episode flattened by an ordinary close is dropped (pinned by an existing test). That is the pside analogue of the #1133 coin fix and is queued as its own slice.

[...]

PR #1133, opened 2026-07-07
HSL: derive replay cooldown/no-restart from canonical RED episodes

enarjord:
## Summary

Closes the remaining gap in the "Coin-HSL episode anchoring redesign" tracker item: **terminal no-restart/cooldown accounting from canonical RED timestamps**.

Live coin-HSL startup replay derived cooldown/no-restart state only from bot-emitted panic order markers (`panic_flatten_events` / panic `pb_order_type` fills). An episode that crossed RED and was flattened by an **ordinary (non-panic) close fill** — including a manual close — was silently reset with no stop accounting (`red_seen_in_episode` was computed every replay row but never consulted), so a restart during an active cooldown, or after a terminal-drawdown episode, resumed trading. The Rust backtest halts at the RED flatten regardless of order type, and the B2.1 contract says cooldown activates for any episode with `red_seen_in_episode`, anchored at the scope-flattening fill.

## Change (src/passivbot_hsl.py, replay row loop)

- An ordinary scope-flattening fill of an episode with `red_seen_in_episode` now enters the **same stop-latch path as a confirmed panic marker**: `stop_ts` = flatten fill timestamp (B2.1 episode end), cooldown = `stop_ts + cooldown_ms`, and `restart_after_red_policy`/no-restart evaluated at that stop via the shared Rust `ehsl::no_restart_triggered` predicate (max(raw, ema) at the flatten sample). Latch payload, pnl-reset boundary, forced `graceful_stop`, and elapsed-cooldown restart behave identically to the panic-marker path.
- RED-free ordinary flattenings keep the existing plain episode reset.
- Reconstruction logs now carry `source=red_episode_flatten` vs `source=panic_fill_flatten`.
- `red_seen_in_episode` comes from the shared Rust runtime (confirmed min(raw, ema) RED semantics), so the decision predicate stays centralized in Rust; the Python change is replay/reconciliation orchestration, per the centralization guiding contract.

## Tests

Five new replay tests (all four latch scenarios **fail without the fix**; the control passes):
- cooldown latch anchored at the flatten fill, active at boot → halted + `graceful_stop`;
- terminal no-restart when the at-flatten drawdown breaches the threshold (policy=threshold);
- terminal no-restart under policy=never on any RED episode;
- RED recovered before the flatten still activates cooldown (red_seen drives it; no-restart evaluated on recovered at-flatten drawdown → not terminal);
- RED-free ordinary flatten keeps the plain reset (no stop state).

[...]

PR #1131, opened 2026-07-07
Anchor HSL cooldown at the scope-flattening fill

enarjord:
Third behavior slice of the #1122 episode/RED contract (B2.1 item 4): **cooldown anchors at the fill that flattened the scope, by any means** — not at the latest bot-emitted panic fill. Previously, a position finished off manually (or by any non-panic close) after the last panic fill anchored cooldown at the earlier panic fill, starting — and therefore expiring — the window prematurely.

Scope:
- New `_equity_hard_stop_latest_flatten_fill_timestamp_ms` (+ optional variant): the scope's latest fill of any type within the window — once the scope is flat, that is by definition the flattening fill. Layered fallback keeps anchors monotone on missing data: no fill evidence in window → the previous panic-fill anchor → the caller fallback (exchange now).
- All five live cooldown anchor sites switched: both RED supervisors, the check-path finalization for recovered episodes (#1129), and both repanic-refresh paths. For bot-flattened episodes the anchor is unchanged (the panic fill *is* the latest fill); the behavior delta is exactly the manual-completion case.
- Replay-side marker anchoring is deliberately unchanged — episode reconstruction from the canonical series owns that per B2.7.
- Bound on `Passivbot`, pinned in the AST + fake-bot binding tests.

Evidence: regression pins the manual-close-after-panic anchor, window-exclusion fallback, and pair isolation; 213+ passed across the HSL suites; full suite shows only the 3 known pre-existing out-of-scope failures (see #1116); CHANGELOG entry; `git diff --check` clean.

With this, only the incomplete-history startup policy remains from the #1122 contract.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1138, opened 2026-07-08
HSL: derive pside/unified replay cooldown/no-restart from canonical RED episodes

enarjord:
## Summary

Pside/unified analogue of #1133 (queued as a follow-up in the #1137 discussion). The pside/unified startup replay derived cooldown/no-restart evidence only from panic markers — an episode that crossed RED and was flattened by an **ordinary (non-panic) close** was silently dropped, so a restart during an active cooldown (or after a terminal-drawdown episode) resumed trading. The B2.1 contract activates cooldown for any `red_seen_in_episode` episode, anchored at the scope-flattening fill, in **all** scopes; the Rust backtest already behaves this way (parity pinned in #1133 for the pside runtime).

## Changes (`src/passivbot_hsl.py`, `_equity_hard_stop_initialize_from_history` row loop)

- Scope flatness is tracked from the timeline's `is_flat` (unified) / `is_flat_{pside}` (pside) flags; rows without the flags conservatively skip transition detection (no crash on reduced histories).
- A RED-seen ordinary scope-flattening enters the **same stop-latch path** as a confirmed panic marker: `stop_ts` = latest scope fill inside the flatten window (`bisect` over the history's fill timestamps, window bounded by the previous recorded row; falls back to the flatten row minute when no fill evidence exists), cooldown = `stop_ts + cooldown_ms`, and no-restart evaluated via the persistent cross-episode tracker (`_equity_hard_stop_record_no_restart_stop`) exactly like the marker path.
- RED-free ordinary flattenings perform a plain episode reset (`_equity_hard_stop_reset_after_restart`), clearing the episode's RED memory instead of carrying `red_seen_in_episode` into the next episode.
- Reconstruction logs carry `source=red_episode_flatten` vs `source=panic_fill_flatten`.

## Tests

Four-scenario matrix in `tests/test_unstucking_safeguards.py` (the three latch tests **fail on the unmodified replay**; the control passes both ways):
- fill-anchored cooldown latch (`stop_ts` = the scope fill at 195_000, not the row minute);
- terminal no-restart when the stop drawdown breaches the threshold (policy=threshold);
- terminal no-restart under policy=never on any RED episode;
- RED-free ordinary flatten resets without stop accounting.

The #1137 state-parity test's RED-crossing fixture now latches under the new contract; its expectations pin that outcome (parity across row shapes still asserted). The existing `does_not_latch_recovered_red_without_panic_marker` pin still holds — its scope never flattens, and a held recovered-RED episode must indeed not latch.

Full local suite green except the 3 known out-of-scope failures.

@vigilpbclaw-hash @claude @codex please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- HSL pside/unified startup replay now derives cooldown and no-restart
+  evidence from canonical reconstructed RED episodes, matching the coin-mode
+  behavior shipped earlier: an episode that crossed RED and was flattened by
+  an ordinary (non-panic) close now latches its cooldown anchored at the
+  scope-flattening fill (falling back to the flatten row minute when no fill
+  evidence exists) and evaluates restart_after_red_policy/no-restart at that
+  stop via the persistent cross-episode tracker. Previously such episodes
+  were silently dropped, so a restart during an active cooldown resumed
+  trading. RED-free ordinary flattenings now perform a plain episode reset
+  (clearing the episode's RED memory) instead of carrying state into the
+  next episode.
+
 - HSL: new pure, unwired synthesis helper
   `_hsl_replay_pside_timeline_rows_from_cache` converts persisted held-pair
   matrices plus the schema-v5 account series into the aggregate timeline
--- a/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
+++ b/docs/plans/risk_unstuck_hsl_fable_audit_action_plan_20260702.md
@@ -996,6 +996,14 @@ Remaining implementation details:
       `persistent_drawdown_raw` equals the at-stop drawdown ratio, which is
       exactly what the live coin replay and coin forward finalize evaluate.
       With that, every sub-item of this tracker entry is implemented.
+      Implemented (pside/unified parity, 2026-07-08): the pside/unified
+      startup replay now applies the same canonical RED-episode accounting
+      as coin replay - ordinary scope-flattening of a red_seen_in_episode
+      episode latches cooldown at the flatten fill (row-minute fallback) and
+      evaluates no-restart via the persistent tracker; RED-free flattenings
+      reset the episode state. Flatness comes from the timeline's
+      is_flat/is_flat_{pside} flags (rows without the flags conservatively
+      skip transition detection).
 - [ ] HSL replay performance/readiness slice.
       Persist verified non-authoritative HSL time series/checkpoints, add doctor
       tools, prioritize held scopes, keep timing/source evidence, and move dense
--- a/src/passivbot_hsl.py
+++ b/src/passivbot_hsl.py
@@ -4,6 +4,7 @@ import asyncio
 import hashlib
 import json
 import logging
+import bisect
 import math
 import os
 import time
@@ -4418,6 +4419,15 @@ async def _equity_hard_stop_initialize_from_history(self) -> None:
                     state["pending_red_since_ms"] = int(current_metrics["timestamp_ms"])
                 continue
             ignored_panic_marker_timestamps: set[int] = set()
+            scope_flat_key = "is_flat" if signal_mode == "unified" else f"is_flat_{pside}"
+            scope_fill_ts = sorted(
+                _equity_hard_stop_fill_timestamp_ms(fill)
+                for fill in fill_events
+                if signal_mode == "unified"
+                or _equity_hard_stop_fill_pside(fill) == pside
+            )
+            scope_was_nonflat = False
+            prev_recorded_ts: Optional[int] = None
             for row in timeline:
                 if not isinstance(row, dict):
                     continue
@@ -4455,7 +4465,17 @@ async def _equity_hard_stop_initialize_from_history(self) -> None:
                     latch_red=False,
                 )
                 n_rows[pside] += 1
+                row_flat = row.get(scope_flat_key)
+                scope_flattened_this_row = (
+                    isinstance(row_flat, bool) and row_flat and scope_was_nonflat
+                )
+                if isinstance(row_flat, bool):
+                    scope_was_nonflat = not row_flat
+                row_prev_recorded_ts = prev_recorded_ts
+                prev_recorded_ts = ts
                 panic_flatten_marker = panic_flatten_events_by_key.get((pside, ts))
+                stop_ts: Optional[int] = None
+                stop_source = "panic_fill_flatten"
                 if panic_flatten_marker is not None:
                     marker_ts = int(panic_flatten_marker["timestamp"])
                     if not _equity_hard_stop_replay_marker_confirms_red(current_metrics):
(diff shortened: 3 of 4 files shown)


enarjord (commented):
**Reviewer:** Claude (Opus 4.8) · continuous reviewer
**PR:** #1138 "HSL: derive pside/unified replay cooldown/no-restart from canonical RED episodes"
**Head SHA:** `598b67d7` · **Base:** v8 `81881153`
**Scope reviewed:** full diff (4 files) — `src/passivbot_hsl.py` (`_equity_hard_stop_initialize_from_history` row loop, the live pside/unified startup path), tests, CHANGELOG, docs. **MAX-care** (cooldown/no-restart authorization, RED-split fields) per [[reviewer-lesson-authorization-split-misses]].
**Verdict: approve** ✅

The pside/unified twin of #1133 (which I reviewed for the coin scope). Closes the same real risk-safety gap in the fail-safe direction: a pside/unified episode that crossed RED and was flattened by an *ordinary* (non-panic/manual) close was silently dropped → restart during an active cooldown or after a terminal-drawdown episode resumed trading. Now such an episode enters the same stop-latch path as a confirmed panic marker.

**Control flow — verified correct** (`_equity_hard_stop_initialize_from_history`, passivbot_hsl.py:4465+):
- Scope flatness tracked from `is_flat` (unified) / `is_flat_{pside}` (pside); `scope_flattened_this_row = row_flat and scope_was_nonflat`, and `scope_was_nonflat` only updates when the flag is a real bool — reduced histories skip transition detection (no crash), and an initially-flat scope can't spuriously latch (needs a prior non-flat row).
- Panic marker present → `stop_ts = marker_ts` (unchanged, still behind the `_replay_marker_confirms_red` gate).
- `elif scope_flattened_this_row` + `red_seen_in_episode` → `stop_ts` = latest scope fill in the window `(row_prev_recorded_ts, ts+60_000)` via `bisect` over the scope's fill timestamps, fallback to the row minute; `source=red_episode_flatten`. Anchor window is bounded below by the previous recorded row so it can't pick a fill from an earlier episode. Consistent with #1131/#1133 flatten-fill anchoring.
- `elif scope_flattened_this_row` + RED-free → `_equity_hard_stop_reset_after_restart(pside)` + `continue`, explicitly clearing the episode's RED memory so `red_seen_in_episode` can't carry into the next episode (this directly addresses the cross-episode concern I flagged on #1133).
- `if stop_ts is None: continue` correctly gates the shared latch path so only a panic marker or a RED-episode-flatten reaches it. Shared path anchors cooldown at `stop_ts`, evaluates no-restart via the persistent cross-episode tracker (`_equity_hard_stop_record_no_restart_stop`), builds the latch payload with `stop_ts`, logs `source=%s`; terminal → `break`.

**Not dead code / field availability.** `red_seen_in_episode` is produced for pside/unified metrics by `_equity_hard_stop_apply_sample` (passivbot_hsl.py:3234/3382, `bool(step["red_seen_in_episode"])` from the #1128 runtime step) — same machinery the coin path (#1133) consumes. Rust production is unchanged; parity for the recovered-RED→cooldown backtest behavior was already pinned for the pside runtime in #1133.

**Tests ran** (fresh maturin build at `598b67d7`; no Rust production change):
- The four-scenario matrix passes: fill-anchored cooldown (`stop_ts` = the scope fill at 195_000, not the row minute); terminal no-restart on threshold breach; terminal no-restart under policy=never; RED-free ordinary flatten resets without stop accounting.
- `test_unstucking_safeguards.py` + `test_passivbot_balance_split.py` + `test_hsl_coin_mode.py` → **434 passed, 0 failed**. The #1137 state-parity RED-crossing fixture now latches under the new contract (expectations updated, cross-row-shape parity still asserted); `does_not_latch_recovered_red_without_panic_marker` still holds (its scope never flattens).

**Minor (non-blocking), carried over from #1133:** no test exercises *two* episodes in one replay where a RED episode latches a non-terminal cooldown and a subsequent RED-free episode flattens ordinarily — asserting it does not re-latch. The non-terminal cooldown path `continue`s without an explicit `reset_after_restart`, so it relies on the runtime resetting `red_seen_in_episode` at the episode boundary (shared #1128 machinery; the RED-free branch adds an explicit reset on top). A multi-episode replay pin would directly guard this. Worth noting the worst case is fail-safe: a spurious re-latch would be an *extra* cooldown (more conservative), never a missed halt — and missed halts are exactly what this PR removes.

**Nit:** the `if True:` wrapping the shared latch block (kept to minimize the diff after hoisting the `stop_ts is None` guard) reads oddly; a future cleanup could drop it.

No blocking findings. Correct, well-tested, fail-safe, and consistent with the coin arc. CHANGELOG + docs accurately describe it and correctly note this is the pside twin of the #1133 coin fix.

vigilpbclaw-hash:
PR: #1138
Head SHA: 598b67d73cf4ed48475fe56c16c4fcdb0ede8af7
Base SHA/branch: 818811536e60a20d1d9ad29b2d32df2f1550fc38/v8
Scope reviewed: full PR
Verdict: approve

Findings:
- Blocker: none.
- Important: none.
- Minor: none.

Intent/architecture pass:
- Intent inferred: close the pside/unified HSL startup-replay gap where a RED episode flattened by an ordinary close was not reconstructed as a cooldown/no-restart stop, while preserving RED-free flatten resets.
- Design fit: using the existing replay episode state (`red_seen_in_episode`), timeline flatness flags, fill-time anchoring, and the existing latch/no-restart payload path is consistent with the coin-mode contract and avoids introducing a separate risk mechanism. Rows without flatness flags skip transition detection rather than guessing, which is an acceptable conservative compatibility boundary.

Validation performed:
- Dry-run merge into current `origin/v8` (`git merge --no-commit --no-ff origin/pr-1138`) — passed.
- `uv venv --python 3.11 .venv && uv pip install -r requirements.txt -r requirements-dev.txt pytest pytest-asyncio` — review environment prepared.
- `python -m pytest -q tests/test_unstucking_safeguards.py` — passed (120 tests).
- GitHub checks observed: `build` passed.

Notes:
- Reviewed current head `598b67d73cf4ed48475fe56c16c4fcdb0ede8af7`; no prior Hermes Continuous Reviewer comment existed for this PR/head at review start.

enarjord:
Codex review: green at `598b67d73cf4ed48475fe56c16c4fcdb0ede8af7`.

Findings: none blocking. The delta correctly moves pside/unified startup replay from panic-marker-only evidence to canonical RED-episode evidence: once a scoped timeline transitions from non-flat to flat and `red_seen_in_episode` is true, it enters the same latch path as a confirmed panic marker, anchors cooldown at the latest scoped fill in the flatten window when available, and evaluates no-restart through the existing persistent tracker. RED-free ordinary flattening resets the episode state without stop accounting. That matches the stated B2.1 contract and the earlier coin-mode direction.

Validation run from `/private/tmp/passivbot-review-pr1138` at `598b67d7`:
- `git diff --check origin/v8...HEAD` passed.
- `python -m py_compile src/passivbot_hsl.py tests/test_unstucking_safeguards.py` passed.
- Focused new/adjacent tests passed: the 4 new red-episode flatten tests, #1137 pside parity/synthesis tests, active-cooldown reconstruction, replay cooldown reset, no-restart peak preservation, and recovered-RED-without-flatten control.
- Full `tests/test_unstucking_safeguards.py` passed: 120 passed, with only the existing portalocker warning.

Fake-live smoke caveat: I ran `src/tools/run_fake_live.py configs/fake_live_hsl_btc.hjson scenarios/fake_live/hsl_long_red_restart.hjson --snapshot-each-step` because this touches live HSL behavior. It failed after the panic close with `staged planner precondition failed before market snapshot refresh: missing current-epoch surfaces=balance,fills,open_orders,positions`. I then ran the same command on current `origin/v8` (`818811536`) and reproduced the same failure, so I am treating that as pre-existing fake-live/live-loop drift, not a #1138 regression. The scenario also exercises coin-mode forward live HSL rather than the new pside/unified startup replay path.

Non-blocking notes: the `if True:` wrapper around the shared latch block is cosmetic, and a future test for multiple episodes in the same pside/unified replay would be useful. Neither changes the review result because the current PR fixes a missed-cooldown failure in the conservative direction and has direct regression coverage for the behavior it adds.

PR #1389, opened 2026-07-24
Fix trailing confirmation with partial fill history

enarjord:
## Summary

- keep trailing positions fail-closed when the latest fill's reconstructed after-state does not match the authoritative exchange position
- preserve stable content-based identities for id-less fills so loading older history cannot fabricate a position-changing epoch
- preserve `last_refresh_ms` as the timestamp of an actual exchange fetch rather than local cache maintenance
- retry unresolved confirmation from the position/fill anchor with bounded in-memory backoff
- start recovery before the earliest exchange-provided position/open timestamp, even when that predates the configured PnL accounting window
- preserve distinct opening and update timestamps through snapshot normalization; update-only timestamps do not masquerade as opening boundaries
- wait for the required recent post-snapshot confirmation before widening historical recovery
- progressively widen timestamp-free recovery per coin and position side, preserving progress when other cohorts join or leave
- reserve potentially expensive widening for the nonblocking `routine_prefetch:trailing_recovery` path; mandatory account-wide fill confirmations remain recent
- cap recovery by connector pagination capacity: two years for Bybit and one year otherwise, with a one-day retry delay at the bound
- traverse empty recent Bybit trade-history windows so sparse older fills remain reachable
- keep stale fill state in the blocking authoritative refresh instead of displacing it with background recovery
- schedule unresolved historical recovery outside the account-wide execution barrier

## Safety boundary

A mismatched fill remains pending until refreshed history reconstructs a post-fill size and average matching the authoritative exchange position. Price and quantity alone are not accepted as proof of a flat-to-position transition because unseen reductions can produce the same live shape. Partial entries, DCA fills, mismatched averages, unchanged identities, missing anchors, and incomplete post-snapshot refreshes remain nontradable.

Historical recovery is an economy mechanism, not a relaxation: unresolved coin/position-side cohorts remain fail-closed when the venue no longer exposes enough history, while unrelated symbols and order classes remain tradable after the normal authoritative account confirmation settles. Long history pagination cannot be pulled into a mandatory account-wide confirmation.

## Validation

- `python -m compileall -q src tests/test_fill_events_manager.py tests/test_unstucking_safeguards.py tests/test_passivbot_balance_split.py tests/test_order_orchestration.py tests/test_passivbot_monitor.py tests/test_hyperliquid_balance_cache.py tests/test_order_reconciliation_contract.py`
- `python -m pytest -q tests/test_fill_events_manager.py tests/test_unstucking_safeguards.py tests/test_passivbot_balance_split.py tests/test_order_orchestration.py tests/test_passivbot_monitor.py tests/test_hyperliquid_balance_cache.py tests/test_order_reconciliation_contract.py`
- `git diff --check origin/master...HEAD`
- GitHub CI: Python 3.12, Python 3.14, and Rust

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ All notable user-facing changes will be documented in this file.
   fills use stable content-based identities rather than history-list indices. Position snapshots
   preserve distinct exchange opening times, while timestamp-free positions retry with
   progressively wider history windows and retain the fills confirmation barrier between attempts.
+  Widening starts only after the required recent post-snapshot confirmation, tracks progress per
+  coin and position side, and is capped at a two-year live recovery horizon to avoid unbounded
+  exchange pagination.
 
 - Bybit closed-PnL refreshes now cover requested history with explicit,
   contiguous sub-seven-day windows and cursor pagination inside each window.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ All notable user-facing changes will be documented in this file.
   exchange state; price and quantity alone cannot prove a flat-to-position transition. Id-less
   fills use stable content-based identities rather than history-list indices. Position snapshots
   preserve distinct exchange opening times, while timestamp-free positions retry with
-  progressively wider history windows and retain the fills confirmation barrier between attempts.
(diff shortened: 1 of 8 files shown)


enarjord:
@codex review

PR #1393, opened 2026-07-24
Recover missing live fill confirmations after restart

enarjord:
## Summary

- preserve `last_refresh_ms` as the timestamp of an actual exchange fetch instead of advancing it during local cache loading, normalization, or doctor repair
- widen trailing fill-confirmation refreshes back to the authoritative position/fill anchor with bounded in-memory retry backoff when recent overlap cannot confirm the position
- accept a narrowly proven full-opening fill when truncated older cache history polluted reconstructed `psize`/`pprice`
- keep affected trailing positions nontradable until the normal fill-confirmation predicates pass

## Root cause

Two restart recovery paths could leave valid live positions permanently unconfirmed:

1. Loading a local fill cache advanced `last_refresh_ms` before any exchange request, so the first incremental refresh could skip fills which occurred while the bot was offline.
2. A retained history window beginning mid-position could carry a phantom reconstructed position into later fills, making the latest fill's cached after-state disagree with the authoritative exchange position even when that fill exactly opened the whole current position.

The recovery fetch is targeted and RAM-only. It retries from five minutes before the affected position/fill anchor with 5, 10, 20, 40, then 60 minute backoff, capped by the configured fill-history lookback. It does not relax the fail-closed trading gate.

## Validation

- `python -m compileall -q src tests/test_fill_events_manager.py tests/test_unstucking_safeguards.py tests/test_passivbot_balance_split.py`
- `python -m pytest -q tests/test_fill_events_manager.py tests/test_unstucking_safeguards.py tests/test_passivbot_balance_split.py`
- `git diff --check origin/master...HEAD`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable user-facing changes will be documented in this file.
   so extreme optimizer candidates terminate normally instead of panicking an optimizer worker.
   Other invalid orchestrator inputs now propagate as backtest errors without unwinding Rust.
 
+- Live fill confirmation now preserves the last successful exchange-refresh timestamp while
+  loading or repairing local fill caches, widens fill-history fetches with bounded backoff when a
+  position remains tied to a stale or mismatched fill, and recognizes an exact full-opening fill
+  even when older truncated history polluted its reconstructed post-fill size. Affected trailing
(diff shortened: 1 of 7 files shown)


enarjord:
Closing as superseded by #1389, which contains the same recovery contract plus the subsequent review fixes and broader regression coverage.

PR #1397, opened 2026-07-25
Force WEEX CCXT clients to use IPv4

enarjord:
## Summary

- force both WEEX REST and Pro/WebSocket CCXT clients to create IPv4-only `aiohttp` connectors
- preserve CCXT's normal lazy session setup, SSL configuration, throttling, and close ownership
- document the WEEX IPv4 whitelist/`-1056 ILLEGAL_IP` contract and add an Unreleased changelog entry

## Root cause

On a dual-stack host, CCXT/aiohttp could connect to `api-contract.weex.com` over IPv6. WEEX's API-key whitelist UI accepts IPv4 entries, so public endpoints remained healthy while every authenticated endpoint returned `-1056 Invalid IP address`, even though the host's public IPv4 was correctly whitelisted.

IPv6-first retry/fallback was deliberately not added: an IPv6 attempt would be guaranteed to fail for a key whose whitelist cannot represent IPv6. The connector is therefore IPv4-only for WEEX, with no changes to other exchanges.

## Validation

- `python -m pytest -q tests/exchanges/test_weex.py` — 47 passed
- WEEX plus shared CCXT hook, exchange-name, and reconciliation suites — 244 passed
- deployed default `AsyncWeex` authenticated balance probe succeeded with `AF_INET`
- real WEEX live bot completed account/fill initialization, candle warmup, order WebSocket startup, Rust planning, cancel-first reconciliation, and acknowledged order creation with zero runtime errors

The production VPS venv intentionally lacks pytest, so tests were run in the repository development venv; the environment-specific authenticated and live checks were run on the deployment itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- WEEX REST and private WebSocket clients now use IPv4 transport so API keys
+  bound to the host's public IPv4 address do not fail with `-1056 ILLEGAL_IP`
+  when a dual-stack host would otherwise select IPv6.
+
 - Documented Python 3.12 and 3.14 support, explicitly excluding Python 3.13 until its dependency
   set is supported and bounding package metadata to those validated minor versions, and fixed
   reentrant candle fetch-lock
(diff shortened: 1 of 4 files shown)


enarjord:
The WEEX IPv4 transport fix is validated by targeted/shared suites, an authenticated deployment probe, and a real live bot run with successful planning and order reconciliation. @codex review

PR #1412, opened 2026-07-27
Fix forager EMA degradation with resting entries

enarjord:
## Summary

- keep flat, dynamically selected forager symbols on the symbol-scoped EMA degradation path even while an entry order is resting
- mark a symbol nontradable only when its required EMA inputs remain genuinely unavailable, so Rust removes ideal orders and reconciliation retires stale entries without an account-wide restart loop
- preserve entry-cancellation authorization across failed or ambiguous attempts and later EMA recovery only for exact exchange/client order IDs observed while the side was fully managed; orders first appearing after a transition to `manual` or ordinary `tp_only` remain protected
- preserve previously proven dynamic-forager eligibility across Rust's symbol-level nontradable/manual stop independently of side-scoped cancellation authorization; explicit operator `manual` or ordinary `tp_only` ownership transfer revokes that eligibility
- retain flat-symbol degradation before an eligible forager candidate enters `active_symbols`, or when a temporary safety mode such as HSL `graceful_stop`, `panic`, or `tp_only_with_active_entry_cancellation` replaces `normal`; preserve the raw active-cancellation distinction even though Rust normalizes it to tp-only, while explicit fixed/forced `normal`, `manual`, and ordinary `tp_only` contracts remain unchanged
- keep bounded open-ended 1m gaps eligible for provisional in-memory EMA projection through `live.max_active_candle_tail_gap_minutes` (default 10), even when retry metadata records that same missing tail
- provisionally bridge unresolved gaps already bounded by later authoritative candles for current live strategy EMA reads only when the full still-uncovered contiguous gap is no wider than `live.max_active_candle_tail_gap_minutes`; recovered rows split stale broad retry metadata into its actual residual gaps, ordinary and cache-only forager ranking consumers still see unresolved gaps, no provisional row is persisted, and late real candles invalidate affected EMA caches
- keep forager quote-volume and log-range ranking reads strict and cache-separated even for active symbols; coincident strategy log-range spans may use the approved provisional policy without leaking into `forager_m1`
- keep budget-derived forager ranking staleness at least as long as the active-tail grace period (10 minutes by default), while retaining an explicit lower `max_forager_candle_staleness_minutes` as an operator override
- apply the same bounded open-tail policy to stock perps instead of retaining an unbounded venue-category exception
- retain strict required-input behavior for held positions and fixed/explicitly normal sides; ranking degradation cancels only the affected forager side, and forager quote-volume/log-range retain their narrower cache carry-forward contract

## Root cause

The candle health layer correctly classified a short missing suffix as an allowed open-tail projection, but projection rejected matching unresolved retry metadata. A concurrent refresh could also advance the authoritative tail past an older projection anchor while leaving a bounded internal hole, causing sparse rows to be treated as consecutive EMA samples. Separately, EMA degradation changed a forager-managed side to the configured manual stop mode before reconciliation, so the general manual ownership filter could suppress the required stale-entry cancellation or its retry. A side-scoped ranking failure could also make the whole Rust symbol nontradable; after the affected-side order disappeared, a remaining unaffected-side entry could make the next cycle forget previously proven dynamic-forager eligibility and escalate the same recoverable ranking gap to an account-wide error. Startup HSL replay could temporarily replace `normal` with `graceful_stop`, causing a flat bot-owned resting entry to be misclassified as requiring account-fatal EMA readiness. The refresh-budget-derived forager age could also collapse to one minute, prematurely making otherwise recoverable candidates unavailable. Finally, active-symbol quote-volume/log-range helper reads could opt into provisional internal gaps and feed those values into the separate forager ranking bundle.

## Validation

- full `pytest -q` repository suite passes locally on a freshly rebuilt and stamped current Rust extension
- focused candle, EMA degradation, ranking-isolation, and reconciliation suites pass after the final review fixes
- Python compile and `git diff --check`
- regression coverage includes:
  - recorded tail gaps through the exact 10-minute boundary and 11-minute rejection
  - missing authoritative anchor rejection
  - bounded internal provisional gaps, query-clipped long-gap rejection, stale-metadata residual-gap recovery, ordinary/cached-forager isolation, no persistence, and late real-candle replacement/invalidation
  - strict active-symbol forager quote-volume/log-range reads separated from provisional strategy log-range values
  - stock-perp enforcement of the universal open-tail policy
  - mixed fixed/forager side strictness and disabled-opposite-side handling
  - exact-order manual-stop cancellation filtering, failed-cancel retry through later EMA recovery, explicit ownership transfer, side-scoped ranking-EMA degradation, and consecutive degraded cycles with only the unaffected-side resting entry remaining
  - pre-active forager candidates, temporary HSL graceful-stop, and active-entry-cancellation classification for flat bot-owned resting entries
- exact-head live validation at `064ea90dc`:
  - ten existing live bot processes across supported exchange connectors reached market-ready
  - all observed settled health records reported `err=0/10`
  - normal order execution succeeded where actions were due
  - a held-position trailing surface which was initially incomplete recovered within the first completed cycle and Rust evaluated its close threshold
  - unavailable flat candidates remained quarantined without consuming the account-wide crash budget
  - processes remain running on the PR head for overnight observation


enarjord:
@codex review

enarjord:
@codex review

enarjord:
Addressed the latest review findings at `416975257`.

- Internal provisional gaps are now bounded, but deliberately by the live
  `max_active_candle_tail_gap_minutes` policy rather than the review's suggested
  `backtest.gap_tolerance_ohlcvs_minutes`. Importing the simulation-only setting into live
  readiness would violate the existing contract and could permit 120-minute fabricated live
  continuity.
- Cache-only forager quote-volume/log-range carry-forward does not project unresolved internal
  gaps. Current live strategy EMA reads retain the separately reviewed, bounded, non-persistent
  projection and late-real replacement behavior.
- Degraded manual-mode cancellation is authorized only for the exact exchange/client order ID
  observed before degradation; coexisting or replacement operator orders are preserved.
- Required forager ranking degradation carries its position side through cancellation
  authorization.
- The budget-derived forager freshness window now has the 10-minute active-tail grace floor that
  the live contract requires.

The full repository suite, focused regression suites, compile check, and diff check pass.

@codex review

enarjord:
Addressed the current review at `2e693cace`.

Accepted and fixed:

- failed/ambiguous exact-order cancellation authorization now survives later EMA recovery while the same order remains visible, but is dropped on an explicit transition to operator-owned `manual` or ordinary `tp_only`;
- provisional eligibility now measures the full recorded unresolved gap, not only the portion overlapping the requested EMA window;
- cache-only quote-volume/log-range ranking reads remain strict, while bounded provisional close EMAs keep their existing live contract;
- raw `tp_only_with_active_entry_cancellation` is recognized before orchestrator normalization erases its distinction from ordinary `tp_only`.

One finding was deliberately not implemented: an entry observed while a coin+pside is fully managed is bot-owned under Passivbot's reconciliation contract, regardless of whether the user submitted it. Passivbot is required to recreate a user-cancelled desired order and cancel an undesired user-created order unless that coin+pside is `manual` or, for entries, ordinary `tp_only`. Excluding a “coexisting operator entry” that was already visible before EMA degradation would therefore weaken managed-side reconciliation and leave an undesired entry live without its required EMA. Exact order IDs instead provide the intended boundary: entries observed before degradation remain cancellable; orders first appearing after ownership transfers to `manual`/ordinary `tp_only` are protected.

The PR body and AI contract documentation now state this distinction explicitly. Focused regressions and the full repository suite pass at the new head `2e693cace`.

@codex review

enarjord:
Addressed the latest review at `6d90aab0c`.

Accepted and fixed:

- completed-candle forager ranking now uses strict, policy-separated EMA cache keys and cannot
  reuse provisional active-strategy quote-volume or log-range values;
- cache-only candidate close EMA reads explicitly disable provisional internal gaps and likewise
  cannot reuse a provisional active-strategy close cache entry;
- failed-cancellation authorization now preserves both client-order and exchange-order identity
  aliases as an order gains its exchange ID;
- synthetic replacement tracking now uses the candle manager's live/replay clock, preserving
  deterministic invalidation during historical fake-live runs.

One P2 suggestion was deliberately not implemented. Required strategy EMA degradation is
documented and implemented as symbol-scoped: Python supplies symbol-level close/volatility maps,
and Rust receives one symbol-level `tradable` flag. The side-aware safety boundary is that every
normal side must be dynamically forager-managed before the flat symbol may degrade; any fixed or
explicitly normal side keeps the strict failure contract. Missing ranking features are different:
their affected position side is authoritative and already carried separately, so ranking
degradation remains side-scoped. Turning only one side off for a missing strategy span would require
a new partial-bundle/Rust-input contract, not a narrow correction, and would contradict the current
`docs/ai/features/candlestick_manager.md` contract.

The focused suites and the full repository `pytest -q` suite pass, along with Python compile and
`git diff --check`.

@codex review

enarjord:
Addressed the latest valid P1 at `530cbb6d2`.

- Active-symbol quote-volume ranking reads now explicitly reject provisional internal gaps.
- Active strategy log-range and forager ranking log-range are separate reads: the strategy input may retain the reviewed bounded provisional policy, while `forager_m1.log_range` is strict.
- Strict and provisional derived-EMA results have policy-separated cache keys, so a coincident span cannot reuse the wrong consumer's value.
- The strict ranking path retains the existing bounded cached-real carry-forward fallback; it does not manufacture zero-volume or zero-range history.
- Regression coverage proves an open-tail strategy projection can coexist with different cached-real forager values, and verifies that ranking helpers receive the strict policy.

I did not implement the earlier side-specific strategy-bundle suggestion. That remains incompatible with the current symbol-scoped Python/Rust EMA readiness contract; ranking-feature failures remain side-scoped where the affected side is authoritative.

Validation at this head:

- full repository `pytest -q` passes
- focused candle, EMA degradation, ranking-isolation, and reconciliation suites pass
- Python compile and `git diff --check` pass
- ten existing live bot processes across supported connectors reached market-ready on the exact head
- all observed settled health records reported `err=0/10`; normal executions succeeded where due
- a held-position trailing surface initially unavailable at startup recovered within the first completed cycle and Rust evaluated its close threshold

The live processes remain on this PR head for overnight observation. The PR is intentionally left unmerged pending review.

@codex review

enarjord:
The delayed review of `90eae1e2a6` exposed one additional valid P1, now fixed at `064ea90dc`.

The reviewer was correct that dynamic-management eligibility and cancellation permission had become coupled. When only the long ranking feature was missing, Rust's symbol-level `tradable=false` could move both sides to manual while cancellation correctly remained long-only. If only a short entry remained visible on the next cycle, neither side proved dynamic through the old order-ID path and the same ranking gap became account-fatal.

The fix deliberately separates the two contracts:

- previously proven dynamic-forager eligibility is retained in RAM across Rust's symbol-level manual stop;
- eligibility remains constrained to enabled forager sides;
- an explicit operator `manual` or ordinary `tp_only` override revokes it;
- cancellation authorization remains based on the affected ranking side and exact order identities, so the unaffected short entry is not cancelled.

Regression coverage reproduces two consecutive cycles, removes the affected-side order after cycle one, leaves only the unaffected-side order, and proves cycle two remains degraded with no cancellation authorization.

Validation:

- focused consecutive-cycle and ownership tests pass
- full repository `pytest -q` passes
- Python compile and `git diff --check` pass
- all ten live bot processes were restarted on exact head `064ea90dc`
- every process reached market-ready; all observed settled health records report `err=0/10`
- held-position trailing recovered within the first completed TradFi cycle and Rust evaluated its close threshold
- normal live order execution succeeded where actions were due

The bots remain on this head for overnight observation. The PR remains unmerged.

@codex review

PR #1428, opened 2026-07-29
Order same-millisecond fills by the exchange position chain

rylos:
## Problem

Fills that share a single millisecond carry no execution order in exchange responses, and the cache preserves whatever order the response had (every sort in `fill_events_manager.py` keys on `timestamp` alone). Position reconstruction replays fills in list order, and `_latest_fill_position_change_anchors()` treats the last listed fill as the account's latest state.

On a live Hyperliquid account this deadlocked trailing for over a day. Two fills executed in the same millisecond:

| tid | dir | sz | startPosition | after |
|---|---|---|---|---|
| 952357507764053 | Close Long | 0.19 | 653.21 | 653.02 |
| 1109260071634171 | Open Long | 2.04 | 653.02 | **655.06** |

They were cached in reverse order, so the newest anchor was the close, with an after-state of 653.02 while the exchange position was 655.06 @ 56.2405. Every cycle logged:

```
WARNING [hyperliquid] [trailing] trailing state unavailable reason=position_fill_confirmation_pending
  symbols=HYPE action=mark_nontradable_until_fresh
  confirmation_details=HYPE/USDC:USDC:long(failed=fill_after_state_mismatch,fill_ts=1785241167526,
  position_update_ts=None,fill_generation=4647,min_fill_generation=3051)
```

The gap is 2.04 contracts, far outside the qty tolerance, so #1409's price tolerance does not cover it. Hyperliquid exposes no position update timestamp, so `_position_open_fill_replay_matches_state()` is unavailable as a fallback. `mark_nontradable_until_fresh` sets `tradable=false`, so no orders were placed for the symbol at all — and only a new fill could clear the confirmation. The bounded history recovery refetched the same (correctly reported, wrongly ordered) fills and could not resolve it. The account sat with an open position, one stale resting order, and no grid or trailing management for 27 hours.

## Fix

Hyperliquid reports `startPosition` per fill, so a same-millisecond cohort forms an unambiguous chain (`653.21 -> 653.02 -> 655.06`).

- `order_same_timestamp_fills()` reorders each `(symbol, position_side, timestamp)` cohort along that chain before `compute_psize_pprice()`, so replayed `psize`/`pprice` and the persisted cache both end on the true final state. Cohorts without chain evidence, or whose chain is ambiguous (no unique head/successor), keep their existing order.
- `_latest_fill_position_change_anchors()` prefers the chain terminal among fills tied at the newest timestamp, derived from each fill's own recorded after-state and signed qty. This is exchange-agnostic and lets caches already written out of order recover without a rewrite.

Both paths fail open: any ambiguity falls back to today's list order.

## Tests

- `test_latest_fill_anchor_prefers_intra_millisecond_chain_terminal` — reproduces the live cohort; fails on master with `assert 653.02 == 655.06`.
- `test_latest_fill_anchor_keeps_list_order_for_ambiguous_cohort`
- `test_order_same_timestamp_fills_follows_hyperliquid_position_chain`
- `test_order_same_timestamp_fills_ignores_cohorts_without_chain_evidence`
- `test_order_same_timestamp_fills_keeps_distinct_timestamps_untouched`

`tests/test_fill_events_manager.py` and `tests/test_unstucking_safeguards.py` pass in full.


enarjord:
@codex review

PR #1429, opened 2026-07-29
Persist finalized WebSocket candles for flat forager candidates

enarjord:
## Summary

- use CCXT Pro `watchOHLCV` for public 1m candle updates on flat forager candidates
- persist proven-final WebSocket candles through `CandlestickManager`'s canonical 1m shards/index instead of keeping a provisional RAM-only overlay
- let canonical readers, EMA calculations, candle health, and restart recovery consume the same persisted candle state
- retain REST as the complete fallback for startup basis, history, gaps, silence, reconnect recovery, corrections, and forced periodic integrity audits
- dynamically subscribe only approved symbols on position sides actually using forager mode; active-position/open-order symbols continue through the existing urgent REST path

## Finality and parity contract

CCXT Pro normalizes exchange-specific candle messages but may return either fresh updates or a repeated sliding cache. Passivbot therefore keeps a small RAM-only observation ledger per uninterrupted watcher session. The first nonempty snapshot of every watcher session only primes provenance. Thereafter, a changed row may correct a timestamp already present in canonical history. Extending canonical history with a new timestamp requires a fresh successor proving a trusted preceding bucket closed; local processing time is not transport finality evidence.

Additional safeguards:

- the current in-progress minute is always rejected
- REST/disk must first establish a canonical basis; WebSocket data cannot bootstrap trading history by itself
- past rows replayed in the first post-reconnect snapshot remain ineligible for successor sealing until that row itself changes
- a current-bucket update emitted before close but processed after it cannot extend history, whether its values changed or remained unchanged
- WebSocket silence never proves a no-trade minute and never synthesizes a zero candle
- watcher provenance is cleared after an exception/reconnect, so proof cannot bridge a disconnected interval
- large CCXT snapshots are trimmed to the newest three rows; reconnect gaps are recovered by REST
- later changed WebSocket or REST rows overwrite the same timestamp and invalidate ordinary and projected 1m EMA caches
- every WebSocket shard write is read-verified through canonical disk selection before the row or its EMA invalidation becomes visible in RAM
- successful REST omission alone does not delete a validated WebSocket candle; both transports are accepted sources of a good finalized candle
- persisted WebSocket metadata survives restart but does not advance the REST refresh timestamp
- periodic integrity audits explicitly force a bounded REST overlap even when the canonical WS tail is already fresh

## REST-only and degradation behavior

`enable_forager_ws_candles = false`, globally disabled WebSockets, unsupported `watchOHLCV`, or a zero OHLCV network budget retain REST-only maintenance. The lightweight subscription reconciler starts whenever the configured transport is available, even if no side is initially in forager mode, allowing later mode transitions without restart. Repeated watcher or ingestion failures back off at 1/2/4/8 seconds and then enter a five-minute cooldown while the independent REST maintainer continues; the failure streak resets only after ingestion succeeds. Subscriptions retry automatically afterward. Ordinary market silence is handled by canonical staleness/gap logic rather than being mistaken for a transport failure.

## Review findings addressed

Initial review:

- require an authoritative basis before admitting WebSocket rows: implemented and regression-tested
- invalidate ordinary EMA caches when candle content is corrected: implemented and regression-tested
- subscribe only position sides actually in forager mode: implemented and regression-tested
- keep the old candidate-only overlay out of global active health: superseded by removing the overlay; persisted rows are canonical for every reader
- delete overlay rows omitted by REST: superseded by the reviewed equal-authority contract; a validated finalized WebSocket candle is retained unless a later row corrects that timestamp

First follow-up review:

- bypass the fresh-tail shortcut for scheduled REST audits: implemented with an explicit forced-overlap refresh and regression tests at both caller and candle-manager layers
- start the watcher when forager mode becomes active later: implemented by keeping the zero-subscription reconciler alive whenever the transport is configured
- reset retry state only after ingestion succeeds: implemented and tested through the five-minute cooldown threshold
- prime post-reconnect snapshots before persistence: implemented with per-row successor eligibility so replayed partial rows cannot be sealed indirectly

Second follow-up review:

- prevent failed shard writes from exposing restart-sensitive WS rows or invalidating EMA caches only in RAM: implemented with persist-before-exposure ordering and failure regression coverage
- require post-boundary proof before a changed row extends canonical history: implemented while preserving changed-row corrections for timestamps already canonical

Latest review:

- reject unchanged rows whose only apparent finality comes from delayed post-boundary processing: implemented by requiring successor proof for every new canonical timestamp
- detect successful-but-invisible shard saves, including immutable complete legacy days: implemented by reading canonical disk contents back and comparing every WS row before cache/EMA exposure

## Compatibility evidence

Before implementation, unauthenticated public CCXT Pro probes produced completed BTC 1m WebSocket candles matching REST on Binance, Bybit, Bitget, OKX, Gate, KuCoin, Hyperliquid, and WEEX. An illiquid KuCoin pair produced no completed event during the probe window, confirming that silence must remain a missing-data condition with REST fallback. Bitget and WEEX returned large cached snapshots per event, motivating bounded tail ingestion.

## Validation

- branch and target were refreshed before publication; PR remains based on current `origin/master`
- full repository `PYTHONPATH=src ... -m pytest -q` passed after the latest fixes
- focused candle/EMA/config regression suites passed
- Python compilation, AI documentation checks, generated live-event registry check, and `git diff --check` passed

All exchange probes used public unauthenticated market data. This branch did not use account credentials or run a live bot.


enarjord:
@codex review

PR #1438, opened 2026-07-31
Unify fill coverage retry ownership

enarjord:
## Summary

- classify unproven fill-history coverage as a controlled authoritative-state block
- move coverage retry cadence to the execution loop and keep it out of generic restart accounting
- remove Passivbot's duplicate coverage timer, status key, and undocumented retry config knobs
- retain `FillEventsManager` as the canonical owner of coverage and known-gap evidence

## Why

The previous inner coverage backoff returned `False`, but staged refresh did not preserve why. The execution loop therefore treated the deliberate wait as a generic authoritative-refresh failure and could restart the bot after eleven deferred cycles. It also left retry policy split between the manager, Passivbot, and the execution loop.

This change keeps planning fail-closed while making the execution loop the single retry-cadence owner. Coverage retries use the existing 30-to-300-second bounds and never consume restart budget. Pending/degraded PnL retries retain their 1-to-60-second behavior.

## Validation

- rebuilt and fingerprint-verified the current Rust extension
- `PYTHONPATH=src .../python -m pytest -q tests/test_passivbot_balance_split.py tests/test_fill_events_manager.py tests/test_realized_loss_gate.py tests/test_unstucking_safeguards.py tests/test_ai_docs.py`
- `PYTHONPATH=src .../python src/tools/check_ai_docs.py`
- `PYTHONPATH=src .../python src/tools/generate_live_event_registry.py --check`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Prevent unproven fill-history coverage from consuming the generic live restart budget; one execution-loop backoff now owns coverage retries while planning remains fail-closed.
+
 - Live fill-history coverage now has one canonical verdict owned by
   `FillEventsManager`. Refresh, staged readiness, HSL replay, and realized-PnL
   consumers no longer duplicate cache/gap interpretation. Metadata that claims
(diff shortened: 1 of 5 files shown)


enarjord:
@codex review

PR #1440, opened 2026-07-31
Consolidate live PnL planning inputs

enarjord:
## Summary

- stop requesting an account-wide confirmation when a known fill is enriched from pending/synthetic to authoritative PnL
- retain account-wide confirmation for genuinely newly discovered fills
- validate and accumulate realized-PnL history once before deriving the Rust auto-unstuck configuration flag
- remove the duplicate auto-unstuck fill scan and safety assertion

## Why

PnL enrichment changes local accounting evidence but does not represent a new exchange-state transition. Treating it like a new fill caused an unnecessary follow-up fetch of balance, positions, open orders, and fills.

Live planning also read and safety-validated the same realized-PnL events once for unstuck eligibility and immediately again for the cumsum values passed to Rust. The cumsum path is now the single validation point; the following configuration check does not reread fills. Rust remains the source of truth for unstuck emission.

This is the third audit PR and should merge after #1438 and #1439; all three independently target `master`.

## Validation

- rebuilt and fingerprint-verified the current Rust extension from this worktree
- fill, state-refresh, HSL, realized-loss, unstuck, order-orchestration, and documentation suites passed
- regression tests prove enrichment requests no account confirmation while a newly discovered fill still requests all account surfaces
- AI documentation and generated live-event registry checks passed
- Python compilation and `git diff --check` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
-- Stop refetching every account surface when a known fill gains authoritative PnL, and validate realized-PnL history once per Rust planning cycle instead of rescanning it for unstuck eligibility.
+- Stop refetching every account surface when a known fill only gains authoritative PnL, while retaining confirmation for new source identities or structural fill changes; validate realized-PnL history once per Rust planning cycle instead of rescanning it for unstuck eligibility.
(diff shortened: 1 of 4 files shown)


enarjord:
@codex review

enarjord:
@codex review

enarjord:
@codex review

PR #1470, opened 2026-08-08
Refine coin override parameter policy

enarjord:
## Summary

- derive the per-side shared-field allowlist from one canonical parameter policy
- allow per-coin `risk.entry_cooldown_minutes` and `unstuck.ema_gating_enabled`
- retain per-coin `unstuck.loss_allowance_pct`
- remove per-coin `risk.we_excess_allowance_mode`; inline use now fails with a global-config migration message, while complete override files warn and ignore the field
- update v7 migration coverage, reference docs, changelog, and the staged rollout checklist

## Validation

- full Python test suite passed
- focused coin-override, config-pipeline, migration, and complete fake-live suites passed
- Rust extension rebuilt and its loaded source fingerprint matched the current Rust source
- CI-style AI docs and generated live-event registry checks passed
- deterministic offline fake-live restart scenario passed with file and inline cooldown/unstuck EMA-gating overrides
- paired real Binance BTC backtests on public cached data completed for 2024-12-10 through 2025-01-02; the per-coin 1,440-minute cooldown plus unstuck EMA-gate case changed worst BTC drawdown from 0.01371 to 0.00391 and produced the expected material behavior delta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Refine the coin-override policy: allow per-coin
+  `bot.<side>.risk.entry_cooldown_minutes` and
+  `bot.<side>.unstuck.ema_gating_enabled`, retain
+  `bot.<side>.unstuck.loss_allowance_pct`, and stop allowing per-coin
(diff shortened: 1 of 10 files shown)


enarjord:
@codex review

PR #1496, opened 2026-08-16
fix(gpu): harden long-horizon MPS drift validation

enarjord:
## Summary

- reduce long-horizon Metal float32 EMA drift with fused delta-form updates in both supported directional kernels
- keep immediate fail-closed behavior for proxy-front constraint classification disagreements
- retain broad-probe disagreements as rolling constraint-agreement evidence, with the existing drift threshold and an eight-probe minimum
- persist per-limit proxy/exact values and violations for actionable diagnostics

## Root cause

The Metal kernels updated EMAs as two weighted products plus an addition. Over long histories, float32 rounding drift could move duration metrics across optimizer-limit boundaries. Broad drift probes intentionally sample less representative regions, so treating any isolated broad-probe classification disagreement as an unconditional halt made healthy long runs stop despite strong rank agreement. Exact Rust remains authoritative: only exact results enter Pareto/storage.

## Safety

- front-selected classification disagreement still halts immediately
- broad-probe constraint agreement must remain at or above `drift_halt`
- rank-correlation gates remain active
- unsupported strategy/runtime surfaces remain fail-closed
- CPU bot, backtest, and optimizer paths are unchanged

## Validation

- `PYTHONPATH=src ./venv/bin/pytest -q tests/optimization/test_gpu_backend.py tests/optimization/test_gpu_service.py` (69 passed)
- `cargo test --no-default-features` (259 passed)
- `PYTHONPATH=src ./venv/bin/pytest -q tests/optimization/test_gpu_mps.py` (12 passed)
- deterministic 955-day ETH run: 61,440 proxy + 64 exact validations; overall rho 0.9996; broad-probe constraint agreement 31/32; no front mismatch
- fresh 59-day BTC run: 61,440 proxy + 64 exact validations in 42.0s; overall rho 0.9976; probe rho 0.9989; front rho 0.9941; constraint agreement 32/32
- Rust extension fingerprint matched the expected source stamp
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,11 @@ All notable user-facing changes will be documented in this file.
   ordering, and hedge/one-way initial-side semantics. Recursive trailing-martingale grid modes,
   other strategies, suites, multi-coin, HSL, auto-unstuck, collateral,
   minimum-effective-cost filtering, market-order execution, incomplete candle tails, non-inert
-  fixed runtime overrides, and unmodeled risk gates fail closed. Proxy/exact optimizer-limit
-  feasibility disagreement also halts immediately. Existing CPU bot, backtest, and optimizer paths
-  do not import or require the optional PyTorch dependency.
+  fixed runtime overrides, and unmodeled risk gates fail closed. Fused delta-form Metal EMA updates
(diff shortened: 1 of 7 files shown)


enarjord:
@codex review

PR #1497, opened 2026-08-16
fix(gpu): preserve exact candle fill boundaries on MPS

enarjord:
## Summary

- preserve Rust's strict candle/order fill comparisons in the Apple MPS proxy by precomputing integer price-tick crossing boundaries
- pass those boundaries to both EMA-anchor and trailing-martingale Metal kernels
- add regression and kernel-contract coverage, plus user-facing documentation

## Root cause

Market candles are stored as float32 for Metal. An order at the exact decimal tick `3711.77` is represented as `3711.77001953125` in float32. Rust compares the candle high as float64 against the exact decimal-tick order price and fills, while Metal previously compared two equal float32 values and missed the strict `high > order_price` crossing. That delayed later fills and could change optimizer-limit feasibility.

The host now computes the highest tick strictly below each candle high and the highest tick at or below each candle low using float64 tick-price semantics. Metal compares integer order ticks to those boundaries, avoiding float32 equality ambiguity.

The immediate proxy/exact front-mismatch halt remains unchanged and fail-closed.

## Validation

- replayed the reported 956-day ETH trailing-martingale candidate: recovery-days proxy moved from `132.461807` to `273.690287`, versus exact Rust `273.689583`; fill count matches at `2238`
- `13 passed` in the full Apple MPS test file
- `69 passed` in GPU backend/service tests
- `259 passed` from `cargo test --no-default-features`
- BTC throughput check: 61,440 proxy evaluations plus 64 exact validations in 41.6 s; generation throughput approximately 1,487/s; rank correlation 0.99959; constraint agreement 1.0 with zero mismatches

Existing CPU bot, backtest, and optimizer paths are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,10 @@ All notable user-facing changes will be documented in this file.
   fixed runtime overrides, and unmodeled risk gates fail closed. Fused delta-form Metal EMA updates
   reduce long-horizon float32 path drift. Proxy-front optimizer-limit feasibility disagreement
   halts immediately; broad-probe disagreements feed a rolling constraint-agreement gate and persist
-  per-limit proxy/exact diagnostics. Existing CPU bot, backtest, and optimizer paths do not import
-  or require the optional PyTorch dependency.
+  per-limit proxy/exact diagnostics. Strict candle/order crossing comparisons are precomputed as
+  integer price-tick boundaries, preventing float32 Metal prices from missing fills that exact Rust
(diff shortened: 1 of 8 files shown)


enarjord:
@codex review

enarjord:
Addressed the tick-price rounding finding in 6bb86cea0: boundary comparisons now mirror Rust's step-decimal-preserving rounding, with the reported price_step=0.1 regression case. Full MPS tests: 14 passed; GPU backend/service tests: 79 passed; the original 956-day ETH replay remains aligned. @codex review

PR #1500, opened 2026-08-16
GPU: support recursive trailing-martingale entries

enarjord:
## What

- allow zero/negative trailing-martingale entry retracement bounds on the Apple MPS optimizer
- reconstruct Rust's full recursive entry ladder from an immutable generation snapshot
- fill every strictly crossed long/short rung in canonical order, including hedge-mode dual-side runs
- retain sequential one-order staging when entry cooldown is nonzero
- keep recursive close grids fail-closed for the next slice

## Why

Exact Rust builds the full passive entry grid one bar ahead using a fixed balance and simulated position. Recomputing each Metal rung from fee-adjusted live state would drift, so the proxy preserves the generation balance, position, price, time, and touch while applying actual fees and position updates per fill.

## Scope and safety

This is additive to the experimental Apple MPS optimizer only. Exact Rust remains authoritative, only exact validations enter the Pareto store, and all CPU/live/backtest paths remain unchanged. Unsupported recursive close grids still fail closed.

## Validation

- 116 GPU Python tests, including real Apple MPS long, short, and hedge-mode recursive ladder tests
- 259 Rust tests with no default features
- cargo check --tests
- cargo fmt --check and git diff --check
- rebuilt/stamped PyO3 extension with exact source fingerprint verification
- bounded ETH long-only recursive-entry optimizer: 2,048 proxy evaluations, 128 exact validations, completed without halt; final 64-sample rho=0.99995, broad-probe rho=1.0, front rho=0.99963, zero constraint mismatches
- interleaved positive-retracement MPS microbenchmark: branch median 0.750s vs master 0.965s for 1,024 candidates x 30,000 synthetic minutes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,13 @@ All notable user-facing changes will be documented in this file.
   feed feasible, diverse candidates into the unchanged exact Rust backtester; only exact results
   reach the Pareto store, and independent broad-probe rank checks halt on proxy drift. Dual-side
   screening preserves separate indicator/trailing/position state, a shared balance, Rust fill
-  ordering, and hedge/one-way initial-side semantics. Recursive trailing-martingale grid modes,
-  other strategies, suites, multi-coin, HSL, auto-unstuck, collateral,
-  minimum-effective-cost filtering, market-order execution, incomplete candle tails, non-inert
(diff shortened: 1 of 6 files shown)


enarjord:
CI is green and local Apple MPS validation is complete. @codex review this exact head, please.

PR #1503, opened 2026-08-16
feat(gpu): add short-only multicoin EMA MPS support

enarjord:
## Summary

- generalize the Apple MPS multicoin EMA Anchor proxy to support either long-only or short-only runs
- preserve the existing long-only runner and Metal entry point as compatibility wrappers
- route the enabled side's EMA, Forager, and position-count parameters while keeping dual-side multicoin runs fail-closed
- mirror short entry/close fills, PnL, unrealized equity, inventory skew, Forager readiness, and TWEL priority in the shared Metal kernel

This remains purely additive. CPU optimization, backtesting, and live bot behavior are unchanged. Multicoin hedged EMA, multicoin trailing martingale, suites, HSL, and auto-unstuck remain unsupported and fail closed.

## Validation

- `cargo fmt --check`
- `cargo test --offline --no-default-features` — 260 passed
- `cargo check --offline --tests`
- `pytest -q tests/optimization` — full suite passed (platform-only skips in sandbox)
- real Apple MPS directional multicoin shader smoke — long and short passed
- three-coin short-only MPS optimization: 4,096 proxy + 64 exact Rust evaluations, no halt
- sustained five-coin short-only optimization from 2022-05 through 2026-08: 4,096 proxy + 64 exact Rust evaluations, no halt; rank rho 0.794, broad-probe rho 0.721, exact Rust remained authoritative for constraint disagreements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ All notable user-facing changes will be documented in this file.
   remains unsupported by the GPU optimizer backend.
 - Add an experimental, purely additive Apple Silicon MPS optimizer backend for single-coin,
   EMA-anchor and trailing-martingale searches in long-only, short-only, and long+short modes.
-  Long-only multi-coin EMA-anchor searches are also supported for up to 64 coins with shared
-  balance, per-coin market and indicator state, dynamic wallet-exposure allocation, Forager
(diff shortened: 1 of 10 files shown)


enarjord:
@codex review

PR #1518, opened 2026-08-18
GPU: support dual-side multi-coin overrides

enarjord:
## Summary
- allow static per-coin EMA Anchor, entry-cooldown, and wallet-exposure overrides on both enabled hedge-mode sides
- pack exact-last long and short override matrices independently for their Metal runners
- include both effective matrices in the checkpoint/runtime identity
- compose the same contract through dual-side multi-coin suites while keeping disabled sides and unsupported leaves fail-closed

## Validation
- full Python test suite
- cargo test --no-default-features: 260 passed
- Apple MPS integration tests: 38 passed
- real M3 MPS single run: Bybit BTC/ETH/SOL, distinct ETH long/short overrides, 2,048 proxy plus 64 exact evaluations, 24.2s, no parity halt
- real M3 MPS suite: 3-coin base plus 2-coin narrow scenario, same overrides, 2,048 proxy plus 64 exact suite evaluations, 33.5s, no parity halt
- cargo fmt --check
- git diff --check

Exact Rust backtests remain authoritative; this only expands the additive Metal screening path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Added static per-coin EMA Anchor strategy, entry-cooldown, and wallet-exposure overrides to
+  dual-side hedge-mode Apple MPS optimization, including compatible multi-coin suites.
+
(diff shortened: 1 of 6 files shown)


enarjord:
@codex review

PR #1521, opened 2026-08-19
fix(optimize): honor suite exchange restrictions

enarjord:
## Summary
- include a combined dataset's full source-exchange identity when resolving optimizer suite contexts
- keep explicitly restricted scenarios on their requested individual exchange dataset instead of misclassifying them as the whole exchange pool
- prevent a Bybit-only scenario from consuming combined candles selected from another base exchange

## Validation
- focused optimizer suite and GPU backend tests
- full Python pytest suite
- Apple MPS pytest file (38 passed on M3)
- original dual-side two-coin GPU suite with base exchanges binance,bybit and two explicitly Bybit scenarios: 2048 proxy + 64 exact validations, 14.4s, no halt, no CLI exchange workaround

## Note
The first full-suite attempt hit an unrelated UTC-midnight test fixture boundary in test_disk_load_observer_receives_summary_and_is_best_effort; the isolated test and the complete suite both passed after the five-minute boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Fixed optimizer-suite exchange routing so an explicitly restricted scenario uses its requested
+  individual exchange dataset even when only that exchange needed separate materialization. It no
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1525, opened 2026-08-19
feat(optimize): support TM coin overrides on MPS

enarjord:
## Summary

- support static Trailing Martingale coin overrides in single-side and dual-side multi-coin Apple MPS optimization
- pack the 23 canonical per-coin TM strategy leaves plus entry cooldown and wallet exposure into an exact-last Metal buffer
- support distinct scenario-local TM override matrices in compatible optimizer suites and checkpoint identity
- keep unsupported override leaves fail-closed and exact Rust portfolio backtests authoritative

CPU optimization, backtesting, and bot operation are unchanged.

## Validation

- full Python test suite
- 42/42 Apple MPS hardware regression tests
- 261 Rust tests plus `cargo fmt --check` and `cargo check --tests`
- AI docs and generated-event registry checks
- rebuilt Rust extension fingerprint verified against the exact source tree
- offline cached BTC+ETH dual-side override smoke: 512 proxy + 64 exact validations, no halt
- offline cached BTC+ETH two-scenario smoke with distinct override matrices: 512 proxy + 64 exact validations, no halt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,23 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Added static per-coin Trailing Martingale overrides to single-side and dual-side multi-coin Apple
+  MPS optimization and compatible suites. The Metal proxy consumes exact-last per-coin strategy,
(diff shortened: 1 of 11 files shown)


enarjord:
@codex review

PR #1533, opened 2026-08-20
feat(optimize): add EMA multicoin MPS TWEL repair

enarjord:
## Summary

- add EMA Anchor side-wide total-exposure repair to the Apple MPS multicoin proxy for long-only and short-only runs
- model both `reduce_overweight` and `reduce_portfolio`, current eligible-count targeting, deterministic adverse-loss ranking, delisted-position price clamping, and exchange quantity/price minimums
- preserve independently reachable ordinary EMA closes while reserving the protective TWEL reducer and accounting for same-candle fills in canonical order
- allow compatible directional suites while keeping dual-side multicoin repair fail-closed until a shared-balance portfolio kernel exists

Exact Rust backtests remain authoritative through the existing exact-validation, classification, rank, and drift gates.

## Validation

- full repository Python test suite: passed on macOS/M3
- `tests/optimization`: passed on macOS/M3, including the complete real-MPS regression file
- `cargo test --no-default-features`: 262 passed
- `cargo check --tests`: passed
- `cargo check`: passed
- `cargo fmt --all -- --check`: passed
- repository Rust-extension rebuild/stamp verified against source fingerprint `49b4c3ce5a79f64ae21144c58e2b86a9c7664874ac8bbab841c809f4214fcc49`
- real two-coin BTC+ETH long-only GPU optimization: 8 generations, 512 proxy evaluations, 64 exact validations, clean completion
- real two-coin BTC+ETH short-only GPU optimization: 8 generations, 512 proxy evaluations, 64 exact validations, clean completion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,22 +4,29 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Added EMA Anchor side-wide total-exposure repair to multi-coin Apple MPS optimization for
+  long-only, short-only, and compatible suite runs. The Metal proxy models both
+  `reduce_overweight` and `reduce_portfolio`, ranks every open position by projected adverse loss,
+  uses the current eligible-position count and last valid delisted-coin price, and reserves the
+  protective reducer before independently reachable ordinary EMA closes. Exact Rust validation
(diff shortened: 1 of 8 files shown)


enarjord:
@codex review

PR #1537, opened 2026-08-20
feat(optimize): gate TM losses on MPS

enarjord:
## Summary

- enable `live.max_realized_loss_pct < 1.0` for single-coin Trailing Martingale Apple MPS optimization
- use a conservative zero-loss proxy envelope instead of enumerating the recursive 500-rung close ladder every candle
- include maker fees and a 1024-ULP float32 safety margin when classifying clearly profitable proxy closes
- preserve exact Rust reducer fallback: try the next protective reducer, then rebuild the ordinary close ladder
- retain fail-closed rejection for all multicoin realized-loss-gate runs
- document the proxy/exact contract and update regression coverage

Exact Rust backtests remain authoritative for the configured peak-relative rolling allowance. CPU optimization/backtesting paths are unchanged.

## Validation

- full Python test suite
- full real Apple MPS suite: 147 passed
- GPU backend/service suites
- Rust: 262 passed with `cargo test --no-default-features`
- `cargo check --tests`
- `cargo fmt --check`
- AI docs validator: 0 errors (pre-existing size warnings only)
- real M3/Bybit ETH single-coin long TM optimization with `max_realized_loss_pct=0.05`: 2,048 proxy evaluations, 32 exact validations, 5.3 seconds, no drift/constraint halt, clean shutdown

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
-- Added cumulative realized-loss gating to single-coin Apple MPS EMA Anchor screening for long,
-  short, hedge-mode, and one-way runs. Metal tracks a conservative all-history peak-relative
-  realized net-PnL budget, including maker fees, and blocks lossy ordinary or exposure-repair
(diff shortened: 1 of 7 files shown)


enarjord:
@codex review

PR #1547, opened 2026-08-21
feat(optimize): accept finite MPS HSL lookbacks

enarjord:
## Summary\n\n- allow finite live.pnls_max_lookback_days in the existing one-sided single-coin MPS HSL topology\n- retain the Metal all-history candidate-local peak as a conservative envelope over exact Rust rolling peaks\n- document that the proxy may trigger HSL early after an old peak expires, while exact Rust validation and drift gates remain authoritative\n- keep malformed lookback values fail closed through canonical parsing\n\n## Validation\n\n- full Python optimization test suite\n- cargo fmt check\n- cargo check tests\n- cargo test without default features\n- physical M3 MPS EMA Anchor CLI smoke at the default 30-day lookback: 8 generations, 256 proxy evaluations, 24 exact validations\n- physical M3 MPS Trailing Martingale CLI smoke at the default 30-day lookback: 8 generations, 256 proxy evaluations, 24 exact validations\n- both smokes completed without drift or classification halt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Added finite `live.pnls_max_lookback_days` support to the existing one-sided single-coin HSL
+  Apple MPS optimizer path. Metal deliberately retains an all-history candidate-local peak as a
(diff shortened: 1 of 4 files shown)


enarjord:
@codex review

PR #1572, opened 2026-08-21
refactor(gpu): share multicoin Metal primitives

enarjord:
## Summary
- compose one Rust-owned Metal primitive module into both EMA Anchor and Trailing Martingale multicoin kernels
- centralize identical rounding, minimum-quantity, fill-accounting, override, allowance, and delisted-price helpers
- add source-contract assertions that every composed multicoin shader contains exactly one copy of each shared primitive

## Scope
This is a behavior-preserving foundation for the future fused dual-side shared-balance kernel. It does not enable any currently unsupported GPU topology or relax HSL, repair, suite, or CPU behavior gates.

## Validation
- `cargo test --no-default-features` (263 passed)
- `cargo check`
- rebuilt the release Python extension on the local Apple M3
- full physical MPS suite: `tests/optimization/test_gpu_mps.py` (all passed)


enarjord:
@codex review

PR #1573, opened 2026-08-21
feat(gpu): add joint multicoin account state

enarjord:
## Summary
- add one reusable Rust-owned Metal account state for future fused dual-side multicoin kernels
- preserve exact long-before-short fill ordering, intraminute realized peak, directional HSL PnL scopes, shared equity, and shared liquidation cutoff
- add physical M3 conformance comparing both position-side HSL traces against the exact Rust runtime

## Scope
Foundation only. This does not enable dual-side multicoin HSL or relax any optimizer fail-closed gate yet.

## Validation
- `cargo test --no-default-features` (263 passed)
- `cargo check`
- rebuilt the release extension on the local Apple M3
- full physical MPS `tests/optimization/test_gpu_mps.py` (all passed)

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -102,9 +102,18 @@ mod tests {
             "inline float coin_override_or(",
             "inline float allowed_wallet_exposure_limit(",
             "inline float clamped_market_price(",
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1574, opened 2026-08-21
refactor(gpu): centralize multicoin fill accounting

enarjord:
## Summary
- make the shared joint portfolio account the sole balance and realized-PnL owner in both multicoin Metal kernels
- route every EMA Anchor and Trailing Martingale entry/close fill through one accounting primitive
- retain directional realized-PnL scopes needed by a future fused dual-side HSL controller
- add source-contract coverage preventing fill paths from bypassing the shared account

## Scope
Behavior-preserving foundation only. This does not enable dual-side multicoin HSL or relax any optimizer fail-closed gate.

## Validation
- `cargo test --no-default-features` (264 passed)
- `cargo check`
- rebuilt the release extension on Apple M3
- full physical M3 `tests/optimization/test_gpu_mps.py` (all passed)
- `tests/optimization/test_gpu_service.py` and `tests/optimization/test_gpu_backend.py` (all passed)

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -145,6 +145,31 @@ mod tests {
     }
 
     #[test]
+    fn multicoin_kernels_route_every_fill_through_joint_account_state() {
+        for (body, expected_fill_paths) in [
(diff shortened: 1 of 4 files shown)


enarjord:
@codex review

PR #1578, opened 2026-08-21
fix(hsl): require unified account flatness

enarjord:
## Summary
- make exact Rust unified HSL treat both sides as one flatness scope
- prevent either unified controller from halting or starting cooldown while any account position or blocking order remains
- preserve side-local `pside` behavior and per-coin `coin` behavior
- extend the hard-stop flat-confirmation regression with an asymmetric opposite-side position

This restores the canonical and live unified HSL contract before enabling the matching Apple MPS path.

## Validation
- `./cargotest.sh --lib -q`: 265 passed
- focused asymmetric Rust regression: passed
- `cargo check`
- rebuilt and stamped the Python Rust extension from the exact branch head
- `pytest -q tests/test_backtest_analysis.py tests/test_hsl_metric_regression.py`: 40 passed
- `rustfmt --edition 2021 --check src/backtest.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Fixed exact Rust backtests so `unified` HSL confirms the whole account is flat before ending a
+  RED episode. A position or blocking order on either side now prevents both unified controllers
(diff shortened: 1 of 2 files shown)


enarjord:
@codex review

PR #1580, opened 2026-08-21
refactor(gpu): share HSL output reductions

enarjord:
## Summary

- consolidate one-side and per-coin Metal HSL metric serialization behind one aggregate reducer
- add dual-side and dual-side-per-coin writers with the same directional, sum, max, count, and conditional-min semantics
- extend the physical MPS joint-controller probe to verify both dual output paths

## Scope

This is an internal, behavior-preserving foundation for fused dual-multicoin HSL. It does not relax the current dual-multicoin `pside` gate. Existing one-sided kernels now exercise the shared reducer; future fused EMA/TM kernels can use the dual writers without duplicating metric logic.

## Validation

- `./cargotest.sh --lib` — 265 passed
- `cargo check`
- `pytest -q tests/optimization/test_gpu_service.py tests/optimization/test_gpu_backend.py` — 474 passed
- `pytest -q tests/optimization/test_optimize.py` — 229 passed
- physical Apple MPS: `pytest -q tests/optimization/test_gpu_mps.py` — 268 passed
- rebuilt and source-stamped the exact-branch Rust extension before MPS validation

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -85,6 +85,18 @@ mod tests {
             source.matches("inline bool update_dual_side_hsl(").count(),
             1
         );
+        assert_eq!(source.matches("struct HslOutputAggregate").count(), 1);
+        for signature in [
+            "inline HslOutputAggregate init_hsl_output_aggregate(",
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1581, opened 2026-08-21
refactor(gpu): encapsulate EMA multicoin side state

enarjord:
## Summary

- move the production EMA multicoin kernel's complete mutable directional portfolio into `EmaMulticoinSideState`
- preserve the proven one-side candle loop through typed thread-local aliases
- add a physical Metal probe that instantiates long and short states concurrently and proves isolation across positions, orders, selection flags, HSL controllers, per-coin controllers, and blocked masks

## Scope

This is a behavior-preserving prerequisite for a fused dual-multicoin EMA kernel. It does not relax current optimizer gates. The next slice can instantiate two production-shaped side states in one candidate thread and interleave them around the existing shared account/HSL primitives.

## Validation

- `./cargotest.sh --lib` — 265 passed
- `cargo check`
- `pytest -q tests/optimization/test_gpu_service.py tests/optimization/test_gpu_backend.py` — 474 passed
- `pytest -q tests/optimization/test_optimize.py` — 229 passed
- physical Apple MPS: `pytest -q tests/optimization/test_gpu_mps.py` — 269 passed
- rebuilt and source-stamped the exact-branch Rust extension before MPS validation

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -313,6 +313,13 @@ mod tests {
         assert!(source.contains("constant int DAILY_COLS = 9"));
         assert!(source.contains("day_min_balance"));
         assert!(source.contains("constant int SCALAR_COLS = 57"));
+        assert_eq!(source.matches("struct EmaMulticoinSideState").count(), 1);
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1584, opened 2026-08-22
refactor(gpu): extract EMA multicoin fill phase

enarjord:
## Summary
- encapsulate chronological EMA multicoin fill accounting in a shared Metal state
- extract one directional fill pass without changing the proven one-side kernel behavior
- keep balance, realized PnL, recovery, duration, volume, and fill counters in one reusable state
- preserve HSL panic and per-coin HSL fill accounting
- add a physical M3/MPS probe that executes long then short fills against one shared account and chronology

This is the execution boundary needed before the fused dual-side EMA kernel can run both sides before a shared unified/pside HSL controller update.

## Validation
- `cd passivbot-rust && ./cargotest.sh --lib` (265 passed)
- `pytest -q tests/optimization/test_gpu_service.py tests/optimization/test_gpu_backend.py` (474 passed)
- `pytest -q tests/optimization/test_optimize.py` (229 passed)
- physical Apple M3/MPS: `pytest -q tests/optimization/test_gpu_mps.py` (271 passed)

No CPU optimizer, backtester, or live-bot behavior is changed.

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -315,8 +315,11 @@ mod tests {
         assert!(source.contains("constant int SCALAR_COLS = 57"));
         assert_eq!(source.matches("struct EmaMulticoinSideState").count(), 1);
         assert_eq!(source.matches("struct EmaMulticoinSideConfig").count(), 1);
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1589, opened 2026-08-22
feat(gpu): add fused EMA multicoin runner

enarjord:
## Summary

- add a persistent Python MPS runner for the fused dual-side, shared-account EMA Anchor multicoin kernel
- decode the fused 62-scalar output contract, including side-specific entry balance and profit/loss reductions
- reuse the existing batching, buffer, end-step, and profiling lifecycle without exposing service routing yet
- prove unified, pside, and coin HSL modes plus fail-closed invalid topology and auto-unstuck behavior on physical Apple MPS

This is an additive internal runner slice. It does not change optimizer service routing, CLI/config behavior, CPU optimization, or exact Rust authority.

## Validation

- Rust: 265 passed
- GPU service/backend: 474 passed
- optimizer: 229 passed
- physical Apple MPS: 276 passed
- Python compile check and `git diff --check`


enarjord:
@codex review

PR #1592, opened 2026-08-22
refactor(gpu): extract TM multicoin side initialization

enarjord:
## Summary
- decode all immutable Trailing Martingale multi-coin side parameters into one side-config value
- centralize Forager weight normalization, derived coefficients, HSL template decoding, and coin-mode selection
- extract per-side state initialization, including per-coin HSL overrides, recursive close-grid state, selection lifecycle, and EMA/trailing extrema seeds
- keep the existing one-side candle loop and output contract unchanged

## Why this slice
A fused long/short TM kernel must load distinct directional parameter rows and initialize two independent strategy/HSL surfaces before sharing account state. This creates that boundary without widening any supported GPU topology yet.

## Safety
- no CPU optimizer, exact Rust backtest, live bot, or exchange path changes
- no GPU scope gate is widened in this PR

## Validation
- Rust: 265 passed with no default features
- Rust default-feature test compile coverage: passed
- physical Apple MPS matrix: 277 passed
- GPU service/backend/optimizer matrix: 711 passed
- loaded Rust extension source stamp matched the exact worktree fingerprint
- git diff --check passed

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -606,6 +606,20 @@ mod tests {
                 .count(),
             1
         );
+        assert_eq!(
+            source
+                .matches("struct TrailingMartingaleMulticoinSideConfig")
+                .count(),
+            1
+        );
+        assert!(source.contains("load_trailing_martingale_multicoin_side_config("));
+        assert!(source.contains("init_trailing_martingale_multicoin_side_state("));
+        assert!(source.contains(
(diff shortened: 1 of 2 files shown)


enarjord:
@codex review\n\nPlease review exact head c6d1f6b4c. Focus on completeness and index accuracy of the 59-column TM parameter mapping, preservation of per-coin override/HSL initialization, Metal address spaces, and behavior preservation.

PR #1599, opened 2026-08-22
Refactor TM multicoin side selection

enarjord:
## Summary

- extract trailing-martingale multicoin candidate eligibility, volume filtering, normalized scoring, slot selection, and score hysteresis into an explicit side-selection helper
- keep coin-HSL entry blocking and reselection state owned by the directional side state
- add a physical-MPS probe proving side-local long/short rankings and coin-HSL isolation

This is a refactoring prerequisite for the fused trailing-martingale shared-account/HSL runner. It does not widen GPU support and does not change CPU optimization, exact Rust behavior, live trading, or backend routing.

## Validation

- Rust tests: 265 passed
- `cargo check --tests`
- rebuilt extension and verified exact source fingerprint
- focused physical-MPS selection/HSL-isolation probe
- full physical-MPS module: 284 passed
- optimizer/backend CPU regression set passed
- token-normalized extraction audit:
  - `selection_prelude_identical=True`
  - `selection_core_identical=True`

## Review focus

- coin-HSL blocked-mask refresh and reselection triggers
- forager volume filtering and score normalization
- deterministic symbol-index tie breaks
- incumbent/challenger hysteresis semantics

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -632,6 +632,7 @@ mod tests {
         assert!(source.contains("update_tm_multicoin_side_indicators("));
         assert!(source.contains("count_tm_multicoin_tradable_coins("));
         assert!(source.contains("tm_multicoin_side_has_position("));
+        assert!(source.contains("update_tm_multicoin_side_selection("));
         assert!(source.contains(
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1602, opened 2026-08-22
Add fused TM multicoin Metal kernel

enarjord:
## Summary

- add the fused long-plus-short trailing-martingale multicoin Metal kernel
- reuse the previously extracted TM side state, fill, selection, order, and dual-HSL phases over one shared portfolio account
- expose the 62-column fused output contract with directional realized-profit and realized-loss reductions
- fail closed before trading when HSL topology differs between sides or auto-unstuck is enabled by either the template or a coin override

This is the Metal-kernel slice only. It does not widen optimizer routing yet. Exact Rust remains authoritative, and the existing CPU optimizer, backtester, and bot paths are unchanged. Dual-side auto-unstuck remains deliberately unavailable here because exact parity requires one global cross-side least-stuck arbitration; a later slice will implement that rather than silently approximating it.

## Validation

- `cargo test --manifest-path passivbot-rust/Cargo.toml --no-default-features -q` — 265 passed
- `cargo check --manifest-path passivbot-rust/Cargo.toml --tests -q`
- rebuilt and stamped the exact Python/Rust extension from this worktree
- focused physical Apple M3/MPS fused-kernel test passed
- full physical Apple M3/MPS module — 287 passed
- CPU optimizer/backend regression matrix passed
- normalized structural audit: fused shared-account control flow is identical to the reviewed EMA fused kernel after accounting for TM parameter columns and tick-control inputs
- `git diff --check`

### DIFF
--- a/passivbot-rust/src/gpu.rs
+++ b/passivbot-rust/src/gpu.rs
@@ -600,6 +600,7 @@ mod tests {
         assert!(source.contains("apply_coin_hsl_overrides("));
         assert!(source.contains("constant int COIN_COLS = 12"));
         assert!(source.contains("constant int SCALAR_COLS = 57"));
+        assert!(source.contains("constant int FUSED_SCALAR_COLS = 62"));
         assert_eq!(
             source
                 .matches("struct TrailingMartingaleMulticoinSideState")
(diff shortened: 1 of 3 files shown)


enarjord:
@codex review

PR #1603, opened 2026-08-22
Expose fused TM multicoin MPS runner

enarjord:
## Summary

- expose `MpsTrailingMartingaleMulticoinFusedRunner` for the merged shared-account TM Metal kernel
- pack and validate the 118-column long-plus-short parameter matrix and independent 44-column coin-override matrices
- dispatch all TM tick-control buffers and decode the shared 62-column directional output contract
- reuse the existing multicoin buffer, sizing, profiling, truncation, and coin-fill-count lifecycle through narrow shader-library and dispatch hooks
- name the fused output width and decoder as a strategy-neutral multicoin contract shared by EMA Anchor and Trailing Martingale

This slice exposes and tests the runner only. It does not widen optimizer/service routing. Exact Rust remains authoritative, the fused TM kernel still fails closed when auto-unstuck is enabled, and existing CPU bot/backtest/optimizer behavior is unchanged.

## Validation

- rebuilt and stamped the exact Python/Rust extension from this worktree
- focused physical Apple M3/MPS fused-runner contract passed
- direct kernel and runner outputs match for daily equity, fills, directional PnL, alive state, and per-coin fill counts
- full physical Apple M3/MPS module — 287 passed
- CPU optimizer/backend regression matrix passed
- Python syntax checks passed
- `git diff --check`

### DIFF
--- a/src/optimization/gpu/mps_kernel.py
+++ b/src/optimization/gpu/mps_kernel.py
@@ -22,7 +22,7 @@ MPS_MULTICOIN_DAILY_COLS = 9
 MPS_SCALAR_COLS = 32
 MPS_MULTICOIN_SCALAR_COLS = 57
 MPS_DIRECTIONAL_SCALAR_COLS = 62
-MPS_EMA_MULTICOIN_FUSED_SCALAR_COLS = 62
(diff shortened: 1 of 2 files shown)


enarjord:
@codex review

PR #1607, opened 2026-08-23
perf(gpu): specialize one-sided TM Metal kernels

enarjord:
## Summary

- add Rust-owned long-only and short-only Metal source variants for single-coin Trailing Martingale
- select the specialized variant automatically only when exactly one side is enabled and HSL is disabled; retain the generic kernel for dual-side or HSL-enabled runs
- keep exact Rust validation, drift gates, optimizer ordering, and CPU paths unchanged

## Performance and parity

On an Apple M3 with 1,924,615 one-minute candles and a 256-candidate batch, the production proxy path improved from 8.94 eval/s with the generic kernel to 19.94 eval/s with the specialized long-only kernel (2.23x). Generic-versus-specialized probes for both long-only and short-only compared all 69 output tensors across 64 candidates and were bit-exact.

## Validation

- `cargo test --no-default-features` (266 passed)
- `cargo check --tests`
- `PYTHONPATH=src python -m pytest tests/optimization -q`
- `PYTHONPATH=src python src/tools/check_ai_docs.py` (0 errors; existing size warnings only)
- generated live-event registry check and AI-doc tests
- rebuilt and source-fingerprint-verified the PyO3 extension
- bounded real-MPS long-only and short-only output-parity probes
- bounded real-MPS production-path performance benchmark

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Accelerated HSL-disabled, one-sided single-coin Trailing Martingale optimization on Apple MPS
+  with Rust-owned long-only and short-only Metal variants. The backend selects these variants
+  automatically, logs the selected topology, and retains the generic kernel for dual-side or
+  HSL-enabled runs. Exact Rust validation and the existing drift gates remain authoritative.
+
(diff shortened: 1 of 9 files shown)


enarjord:
@codex review

enarjord:
@codex review

PR #1609, opened 2026-08-23
fix(gpu): revalidate repeated proxy fronts

enarjord:
## Summary
- keep single-objective and converged Apple MPS searches running when the current proxy front repeats an already exact-validated candidate
- re-run the actual current-front candidate rather than relabeling an off-front candidate, preserving truthful proxy-front drift evidence
- wait for an in-flight current-front exact job before selecting another validation batch
- retire completed hashes from the in-flight set so exact revalidation is possible

## Safety
Exact Rust remains authoritative. Repeated current-front constraint disagreement continues to feed the proxy-front rolling gate and halts after the configured evidence threshold. Broad/off-front candidates are never relabeled as front evidence, and the existing exact-budget, rolling-window, checkpoint, and resume proofs remain intact. CPU optimization and bot/backtest behavior are unchanged.

## Validation
- full `tests/optimization` suite
- focused repeated/completed/in-flight/no-front selection regressions
- rebuilt and verified the Rust extension against source fingerprint `8015fda2867c40ebd34153fd1d9e93460de4d22c7e770bd4d304e940719d317f`
- AI documentation checks: 0 errors (2 pre-existing size warnings)
- cached Apple M3 single-objective HSL CLI reproducer: completed 71 generations, 1,136 MPS proxy evaluations, and all 142 exact validations in 15.7s; the prior code halted around generation 30 when the one-member front repeated
- `compileall` and `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Fixed Apple MPS optimization halting when a converged or single-objective proxy front repeats an
+  already exact-validated candidate. The backend now revalidates that actual current-front member
+  (or waits for its in-flight exact job) instead of relabeling an off-front candidate, preserving
+  the proxy-front drift gate and exact Rust authority while allowing the search to continue.
+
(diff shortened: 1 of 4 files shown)


enarjord:
@codex review

PR #1621, opened 2026-08-23
feat(gpu): support EMA ordinary market execution

enarjord:
## Summary

- add ordinary market-order promotion to the single-coin EMA Anchor Metal proxy
- preserve next-candle execution, adverse directional slippage, taker fees, executable-touch sizing, and one-way arbitration
- allow compatible suite overrides while keeping Trailing Martingale, multi-coin, HSL, unstuck, exposure repair, realized-loss gating, and minimum-effective-cost combinations fail closed
- document the supported boundary and add changelog coverage

Trailing Grid v7 remains intentionally outside GPU optimizer scope.

## Validation

- cargo test --no-default-features: 267 passed
- cargo check --tests: passed
- focused optimizer/backend/service/metrics/suite tests: 602 passed
- complete physical Apple M3 MPS module: 312 passed
- repository-managed Rust extension rebuild and source fingerprint verification: matched
- physical Apple M3 suite smoke: 512 Metal proxy evaluations plus 32 exact Rust validations, completed cleanly with no drift halt
- git diff --check: passed

PR #1622, opened 2026-08-23
feat(gpu): combine EMA market execution with HSL

enarjord:
## Summary

- enable single-coin EMA Anchor ordinary market execution with all supported HSL signal modes
- enable the existing conservative one-sided minimum-effective-cost filter alongside market execution, including simultaneous HSL use
- keep auto-unstuck, exposure reducers, realized-loss gating, Trailing Martingale, and multi-coin market combinations fail closed
- document the expanded support boundary and retain the explicit Trailing Grid v7 exclusion

This is an orchestration scope promotion only: the Metal execution semantics landed in #1621 and are now directly validated in these combinations.

## Validation

- optimizer/backend/service/metrics/suite tests: 605 passed
- complete physical Apple M3 MPS module: 313 passed
- focused HSL plus ordinary-market physical tests: passed for long and short EMA Anchor
- bounded two-scenario physical M3 suite smoke: 512 proxy evaluations plus 32 exact Rust validations across market+HSL and market+HSL+min-cost, completed with no drift halt
- Python compileall: passed
- git diff --check: passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,15 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Extended single-coin EMA Anchor ordinary market-order optimization on Apple MPS to compatible
+  HSL modes and one-sided minimum-effective-cost filtering. Unstuck, exposure-repair, and
+  realized-loss-gate combinations remain fail closed pending their execution-ordering slice.
+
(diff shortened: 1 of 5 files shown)


enarjord:
@codex review exact head c7c4729e68fddc5ca33dd36b08fccdcbd496a911

PR #1640, opened 2026-08-25
Support asymmetric GPU multicoin side universes

enarjord:
## Summary
- allow non-suite dual-side Apple MPS multi-coin runs to use different effective long and short coin universes
- preserve exact payload `entry_eligible` decisions as side-specific zero-WEL Metal overrides for EMA Anchor and Trailing Martingale
- remove obsolete matching-list validators and stale realized-loss-gate error wording
- document the supported topology and add packing/constructor/backend regressions

## Contract
Exact Rust remains authoritative. Config hydration already removes ignored coins from each side's approved set; payload construction then marks every coin/side pair with `entry_eligible`. The GPU packer now carries false eligibility into the existing zero-WEL override column, which the fused kernels already use for side-specific tradable counts, Forager selection, entry generation, dynamic WEL, and one-way arbitration. No Metal kernel changes or CPU-path changes are required. Suite behavior is unchanged.

## Validation
- `python -m pytest -q tests/optimization`
- physical MPS: `test_mps_fused_multicoin_one_way_arbitrates_each_flat_symbol` (EMA Anchor and Trailing Martingale)
- exact extension rebuilt/stamped and source fingerprint verified
- `cargo test --no-default-features` (267 passed)
- `cargo check --tests`
- Trailing Martingale end-to-end asymmetric smoke: BTC long only, ETH short only, 16 generations / 256 proxy / 32 exact, no safety halt
- EMA Anchor end-to-end asymmetric smoke: BTC long only, ETH short only, 16 generations / 256 proxy / 32 exact, no safety halt

## Notes
This is additive to the experimental Apple MPS backend. CPU optimization, backtesting, and live operation are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Added asymmetric long/short coin universes to non-suite dual-side Apple MPS multi-coin EMA
+  Anchor and Trailing Martingale optimization. The proxy now preserves exact payload `entry_eligible`
+  decisions as side-specific zero-exposure coin overrides, so Forager selection, dynamic wallet
+  exposure, HSL, auto-unstuck, and one-way arbitration exclude the same side-ineligible coins as
+  exact Rust without requiring matching long and short approved or ignored lists.
+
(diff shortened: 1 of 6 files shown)


enarjord:
@codex review

PR #1657, opened 2026-08-26
GPU: support recovery through invalid candles

enarjord:
## Summary

- advance single-coin EMA Anchor and Trailing Martingale equity/recovery sampling through internal all-NaN H/L/C gaps using exact Rust's balance-only semantics
- allow equivalent multi-coin timestamps where every declared coin is all-NaN while preserving infinite, finite-unmodeled, and forced-delist fail-closed gates
- preserve NaN packed multi-coin closes so open-position unrealized PnL is omitted exactly rather than marking the position at zero
- admit the supported all-NaN gap only through the active GPU optimizer operation; persisted config values cannot relax backtest, suite, download, or reproduction validation
- make exact Rust classify the same all-NaN row as non-tradable during mandatory exact validation without mistaking an internal gap for a delist
- preserve gap semantics through coarse candle aggregation by ignoring complete gap minutes inside mixed buckets and retaining all-gap buckets as non-tradable
- remove only the obsolete blanket recovery continuity guard; HSL-enabled coins retain their separate contiguous-candle contract
- document the expanded scope and add physical-MPS coverage for long, short, and fused single-/multi-coin topologies

## Review fixes

- address the original P1 by rejecting finite zero/negative, partially invalid, and float32-unrepresentable candles instead of treating them as balance-only gaps
- address the exact-head P1 by applying that safety gate to every GPU proxy run across the entire declared window, independent of the selected equity/recovery metrics
- address the exact-head P2 by allowing all-NaN H/L/C gap rows through `build_backtest_payload` and the production data-preparation validator
- narrow supported gaps from generic non-finite values to all-NaN rows because exact Rust may fill against infinite highs/lows
- address the next exact-head P1 by narrowing the production-data exception to the GPU backend and teaching exact Rust candle-level validity about all-NaN gaps
- address the next exact-head P2 by rejecting non-positive highs and lows in single-coin GPU inputs and in the Metal validity mask
- add unit coverage for the complete fail-closed boundary and physical-MPS open-position coverage across multi-coin NaN gaps
- address the latest exact-head P1 by separating exact Rust's declared valid range from candle tradability so an open position survives an internal gap without a panic close
- address the latest operation-context P2 by threading an explicit GPU-only gap-admission flag instead of inferring permission from persisted `optimize.backend`
- address the latest aggregation P2 with gap-aware coarse aggregation that ignores only complete all-NaN H/L/C rows and lets partial corruption propagate fail-closed
- address the latest forced-delist P2 by rejecting an all-NaN declared endpoint at both raw-data and single-coin proxy boundaries
- address the next exact-head malformed-aggregation P1 by carrying any finite non-positive or float32-unrepresentable raw price into a fail-closed aggregate sentinel
- address the next exact-head stale-order P1 by clearing directional entry and close intents whenever an in-window candle is invalid
- address the next exact-head first-valid P1 at both raw preparation and post-slice single-/multi-coin service boundaries so EMA state is never seeded from a gap
- address the next exact-head interval P2 by scaling the pre-aggregation forced-delist cutoff to the configured candle interval; post-aggregation proxy gates remain authoritative

## Validation

- physical Apple MPS focused NaN-gap matrix: 12 passed
- physical Apple MPS internal-gap recovery and stale-order regressions: 8 passed
- full `tests/optimization` suite on physical Apple MPS: 1,714 passed after the final review fixes
- affected Python regression suites: 484 passed
- exact Rust internal-gap/delist regression: passed
- `cargo test --no-default-features`: 268 passed
- `cargo check --tests --no-default-features`: passed
- rebuilt Rust extension and verified exact source fingerprint
- AI docs tests/check: passed (two pre-existing word-count warnings)
- `mkdocs build`: passed (two pre-existing release-note link warnings)
- Python compileall and `git diff --check`: passed

`ruff` is not installed in the shared test virtualenv.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,22 @@ All notable user-facing changes will be documented in this file.
   exact Rust's balance-only equity, HSL, exposure, restart, daily, recovery, and elapsed-time
   accounting after tracking begins, but do not activate tracking from tradability seen only before
   the requested-start guard. Validity still blocks fills, orders, and unrealized PnL. A timestep
-  covered by a declared valid range but lacking any finite positive packed float32 H/L/C remains
-  fail-closed.
+  covered by a declared valid range whose raw H/L/C values are all NaN is likewise treated
+  as non-tradable balance-only time. Single-coin kernels now apply the same rule to these internal
+  gaps, and exact Rust now classifies the same row as non-tradable without triggering delist panic
+  handling during mandatory validation. Only the GPU optimizer operation explicitly admits these
+  internal gaps; retaining `optimize.backend: gpu` in a config does not relax normal backtest,
+  suite, download, or reproduction validation. Coarser GPU candle intervals ignore complete NaN
+  minutes inside mixed buckets, preserve all-gap buckets as non-tradable gaps, and retain any
+  malformed non-gap price as a fail-closed aggregate. Gap rows must be strictly internal: the
+  first-valid candle and any forced-delist endpoint remain valid finite prices, with the endpoint
+  cutoff scaled to the prepared candle interval. Directional Metal kernels clear pending orders at
+  a supported gap so they cannot fill stale pre-gap intents afterward. This enables
+  account-equity and strategy-equity recovery
+  metrics on such histories. Finite
+  non-positive, partially invalid, or float32-unrepresentable prices remain fail-closed for GPU
+  screening; HSL-enabled coins still require contiguous candles, and a forced-delist endpoint must
+  remain representable because it supplies an executable close.
 
 - Apple MPS single-coin EMA Anchor and Trailing Martingale optimization now mirrors exact Rust's
   forced-delist close when at least 1,400 prepared candles follow a coin's final valid candle,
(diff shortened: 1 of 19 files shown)


enarjord:
@codex review

Addressed the P1 at new head da26b7227: fully non-finite H/L/C gaps retain exact balance-only semantics, while finite zero/negative, partial-invalid, and float32-unrepresentable recovery inputs now fail closed. Multi-coin packing preserves non-finite closes so open positions omit unrealized PnL instead of using zero. Full physical-MPS optimizer tests and focused open-position regressions are green.

enarjord:
@codex review

Addressed both exact-head findings at 4d738a59d. The proxy safety gate now applies to every run across the full declared candle window, and the production backtest-data validator admits the supported all-NaN H/L/C gap while continuing to reject partial-invalid prices, bad volume, infinities, finite non-positive values, and float32-unrepresentable prices. I narrowed the contract from generic non-finite gaps to all-NaN gaps because exact Rust may fill against infinite highs/lows. Focused boundary tests, the 12-case physical-MPS gap matrix, and the complete physical-MPS optimizer suite are green.

enarjord:
@codex review

enarjord:
@codex review

Addressed both findings at exact head ea053b6a8. The all-NaN production-data exception is now GPU-backend-only, normal CPU backtest/optimizer validation remains strict, and exact Rust independently classifies an all-NaN H/L/C row inside a declared range as non-tradable. Single-coin proxy admission and its Metal validity mask now also reject non-positive highs/lows. Focused Python (197 passed), the new exact-Rust regression, full Rust (268 passed), and the physical-MPS optimizer suite are green.

enarjord:
@codex review

Addressed all four findings from the previous exact-head review on the new head `1c4630830`:

- internal all-NaN gaps are non-tradable but no longer trigger exact Rust delist panic handling;
- gap admission is now explicit operation context, so a persisted GPU backend cannot relax other commands;
- GPU coarse aggregation ignores only complete gap minutes inside mixed buckets and preserves all-gap buckets;
- all-NaN forced-delist endpoints are rejected at raw-data and single-coin proxy boundaries.

The rebuilt extension matches the final Rust fingerprint. The 475 affected Python tests, all 1,710 optimizer tests on physical Apple MPS, the targeted Rust regression, and the prior full 268-test Rust suite are green.

enarjord:
@codex review

Addressed all four findings from the previous exact-head review on the new head `c88d72c32`:

- coarse aggregation now carries finite non-positive and float32-unrepresentable raw prices into a fail-closed aggregate sentinel;
- directional EMA Anchor and Trailing Martingale kernels clear pending entry and close intents on every invalid in-window candle;
- raw preparation and post-slice single-/multi-coin service gates reject an all-NaN first-valid candle so EMA state is never seeded from zero;
- the raw forced-delist tail cutoff now scales to the configured candle interval, while post-aggregation proxy checks remain authoritative.

The rebuilt extension matches the final Rust fingerprint. The 484 affected Python tests, all 1,714 optimizer tests on physical Apple MPS, 8 focused physical-MPS recovery/stale-order regressions, the targeted Rust/source-contract regressions, and the full 268-test Rust suite are green.

PR #1663, opened 2026-08-26
Fix equity-balance analysis alignment after warmup

enarjord:
## Summary
- align realized balance changes to equity samples by timestamp when canonical timestamps are available
- preserve the fill-index fallback for timestamp-less analysis inputs
- add a regression covering absolute fill indices after an equity-tracking warmup

## Why
Backtests store absolute candle indices on fills, while the tracked equity vector starts at zero after warmup. Comparing those two indices left stale balances in equity-vs-balance and paper-loss metrics. This was exposed while adding the corresponding opt-in Apple MPS proxy metrics.

## Validation
- cargo test --no-default-features (270 passed)
- cargo check --tests --no-default-features
- git diff --check

Note: cargo fmt --check reports existing formatting drift in passivbot-rust/src/gpu.rs; this PR does not touch that file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Backtest analysis now aligns realized balance changes with tracked equity samples by timestamp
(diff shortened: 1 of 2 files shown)


enarjord:
@codex review

PR #1676, opened 2026-08-28
Reduce default GPU population to 1024

enarjord:
## Summary

- reduce the default Apple MPS NSGA-II population from 4096 to 1024
- retain the independent 4096-candidate dispatch-batch upper bound and support explicit larger populations
- document the latency tradeoff and update default-contract tests

## Why

Long-history GPU runs previously waited for all 4096 proxy evaluations before starting exact Rust validation. A 1024 default reaches the first exact-validation batch four times sooner without changing proxy semantics, exact Rust authority, or the work performed per candidate.

## Validation

- `PYTHONPATH=src python -m pytest tests/optimization/test_gpu_backend.py tests/optimization/test_backends.py tests/test_config_utils_helpers.py -q` (769 passed)
- `git diff --check`
- bounded long-history, one-sided HSL MPS acceptance: default resolved to 1024, one generation completed in seven safety-bounded dispatches, exact validation began, and interrupt between dispatches shut down cleanly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- The Apple MPS optimizer now defaults to a 1024-candidate NSGA-II population instead of 4096,
+  allowing long-history runs to reach exact Rust validation four times sooner. Explicit population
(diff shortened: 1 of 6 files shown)


enarjord:
@codex review

PR #1677, opened 2026-08-28
Specialize inactive Trailing Martingale GPU paths

enarjord:
## Summary

- select dispatch-proven single-coin Trailing Martingale Metal variants from the packed candidate rows
- compile out recursive entry/close grids only when every enabled-side candidate uses positive retracement
- compile out WEL/TWEL enforcers and auto-unstuck only when every enabled-side candidate disables all three reducers
- specialize fixed-off ordinary market execution and realized-loss gating
- ignore HSL configured only on an exposure-disabled side when selecting the active one-side shader topology
- expose the selected dispatch feature flags in opt-in GPU profiles

Mixed or active-feature dispatches retain the full paths. Exact Rust validation, CPU optimization, backtests, and live behavior are unchanged.

## Deterministic Apple MPS benchmark

Fixed seed 7, 512 candidates, 525,600 synthetic one-minute candles, one long side, one dispatch, warm p50 of three runs:

| Case | Base candidates/s | PR candidates/s | Speedup | Base wall | PR wall |
| --- | ---: | ---: | ---: | ---: | ---: |
| `tm-single-long` | 96.37 | 191.01 | 1.98x | 5.313s | 2.680s |
| `tm-single-long-hsl` | 20.58 | 134.47 | 6.53x | 24.88s | 3.807s |

The no-HSL result also matches the initial TM MPS kernel on the identical fixture (190.76 candidates/s in-kernel at `5258a9ac6`) while retaining the current proxy contract.

## Validation

- `cargo test --no-default-features` (275 passed)
- `cargo check --tests`
- repository-owned release extension rebuild and source-fingerprint verification
- complete `tests/optimization/test_gpu_mps.py` real-MPS suite
- `tests/optimization/test_gpu_service.py`, `tests/optimization/test_gpu_proxy_benchmark.py`, and `tests/optimization/test_gpu_backend.py`
- CPU optimizer suite plus the torch/MPS import-boundary regression
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Apple MPS single-coin Trailing Martingale screening now selects dispatch-proven Metal variants
+  that compile out recursive entry/close grids when every active candidate uses positive
+  retracement, and compile out WEL/TWEL enforcers plus auto-unstuck when every active candidate has
+  those reducers disabled. Ordinary market execution and the realized-loss gate are likewise
+  removed only when their fixed run settings disable them. Mixed or enabled-feature dispatches
+  keep the full kernel. HSL configured only on an exposure-disabled side no longer promotes the
+  active side to an HSL kernel. Exact Rust validation remains unchanged and authoritative.
+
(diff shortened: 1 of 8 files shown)


enarjord:
@codex review

PR #1679, opened 2026-08-28
Skip unused TM volatility state on Apple MPS

enarjord:
## Summary

- prove dispatch-wide zero volatility weights before selecting a Trailing Martingale Metal library
- compile out 1-minute and 1-hour volatility EMA state, candle-range loads, and volatility-weight arithmetic only for all-zero active dispatches
- retain the full kernel for mixed, nonzero, or non-finite weights
- extend opt-in GPU profiles with terminated-candidate and estimated post-termination candidate-bar evidence

Exact Rust validation remains authoritative. CPU optimization, backtests, live behavior, and unprofiled host transfer behavior are unchanged.

## Performance

Deterministic in-memory tm-single-long-hsl fixture, seed 7, 512 candidates, 250,000 candles, one 128,000,000 candidate-bar dispatch:

- held-generic versus specialized alternating comparison: kernel p50 1.785 s to 1.698 s, 4.9% faster
- held-generic versus specialized wall p50: 1.821 s to 1.738 s, 4.6% faster
- immutable head 494778bff standard harness: 296.4 candidates/s warm p50, 1.676 s kernel p50, one dispatch

The fixture is public and reproducible with:

    python -m tools.gpu_proxy_benchmark --case tm-single-long-hsl --candidates 512 --dispatch-batch-size 512 --single-bars 250000 --warm-runs 5 --seed 7 --compact

## Validation

- 275 Rust tests with cargo test --no-default-features
- full GPU service tests plus focused dispatch-specialization tests
- physical Apple MPS bit-for-bit comparison against the generic volatility path for long/short HSL, long/short no-HSL, and dual-side kernels
- focused HSL lifecycle, market-entry, open-tail, and CPU import/preflight regressions
- repository-standard release extension rebuild and current-source verification for all five shader exports
- documentation build completed except for the two unchanged strict-mode CHANGELOG link warnings in existing release-note pages
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Apple MPS single-coin Trailing Martingale screening now compiles out the 1-minute and 1-hour
+  volatility EMA state, candle-range loads, and volatility-weight arithmetic when every active
+  candidate in a dispatch uses zero volatility weights for entry and close thresholds and
+  retracements. Mixed or nonzero-weight dispatches retain the full kernel. Opt-in GPU profiles now
+  also report terminated-candidate counts and estimated post-termination candidate-bars, making
+  future early-exit work measurable. Exact Rust validation, CPU optimization, backtests, and live
+  behavior are unchanged.
+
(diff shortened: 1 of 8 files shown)


enarjord:
@codex review

PR #1721, opened 2026-09-09
Add opt-in coupled strategy and unstuck EMA search

enarjord:
Add `couple_unstuck_ema_spans` to `optimize.enable_overrides` to restore the former coupled EMA search. Each candidate's effective strategy spans determine its unstuck spans on CPU and Apple MPS, including coin strategy pins, fixed runtime overrides, mirroring and scenario overrides. Independent search remains the default.

The redundant unstuck EMA genes are omitted. Saved candidates contain explicit spans, including resolved scenario/coin overrides, so ordinary live/backtest replay needs no optimizer hook. GPU packing keeps inherited horizons dependent on each candidate while preserving actual strategy coin pins. Evaluation contracts record the coupling mode without treating derived coin values as fixed policy. Invalid strategy spans fail explicitly. Migration warnings, CLI help and user/architecture docs explain the option and the need for a fresh search when changing modes.

Validation: 1,688 optimizer/configuration/documentation tests; dedicated coverage for all CPU strategies, GPU packing for both supported strategies and sides, anchors, fixed/mirror ordering, warmup, scenario files and override clears, candidate reload and evaluation identity. Real cached-data CLI runs completed with four CPU evaluations and 128 GPU proxy / 64 exact evaluations; saved candidate pairs were verified. The CLI case produced tied zero scores, so it validates execution and persistence rather than strategy performance. GPU EMA/gate behavior and Rust parity are covered by merged #1720.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable user-facing changes will be documented in this file.
 
 ## Unreleased
 
+- Add `couple_unstuck_ema_spans` to `optimize.enable_overrides` to search strategy and unstuck
+  EMA spans together on CPU or Apple MPS, including effective coin/scenario strategy overrides.
+  Coupled search omits redundant unstuck span genes and saves explicit horizons for ordinary
+  replay. Independent search remains the default; changing the option requires a fresh run.
+
 - Auto-unstuck now owns independent `bot.<side>.unstuck.ema_span_0` / `ema_span_1`, including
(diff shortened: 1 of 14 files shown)


enarjord:
@codex review

Please review the current head, including the saved-scenario normalization follow-up. Nested overrides and legacy aliases now normalize before derived spans are saved, and scenario coin patches resolve before coupling. The 932-case focused optimizer/replay/GPU-backend regression run passes.

PR #1733, opened 2026-09-12
Reduce host CPU usage during long CUDA screening waits

enarjord:
CUDA screening can spend a CPU core waiting for a long-running kernel, competing with exact optimizer workers. Record completion on the current CUDA stream; preserve active waiting for up to 0.5 seconds, then yield between event queries before decoding results. The initial active window avoids wake-up overhead for short screening kernels. Keep the existing device-wide synchronization contract and Apple synchronization behavior.

No shader, precision, launch, or strategy calculations change. GPU failures still propagate. Regression coverage checks event polling, asynchronous errors, current-stream completion, the device-wide barrier, and Apple routing.

Validation: 1,875 NVIDIA GPU/optimizer regression tests passed, 77 skipped. Paired fixed-input replay retained exact output equality while substantially reducing host CPU time. Apple hardware validation was not rerun for this change; its routing is covered by tests.

Four paired synthetic EMA/TM, single/multi-coin benchmark cases retained exact outputs and showed less than 0.3% wall-time difference. A bounded optimizer CLI integration also completed successfully on the final implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ since the latest release tag; these features may already be available when insta
 
 ## Unreleased
 
+- Reduce CPU usage during long NVIDIA GPU optimizer screening waits,
+  preserving GPU calculations and Apple GPU synchronization behavior.
+
 - Reduce NVIDIA multi-coin GPU optimizer memory and screening overhead by compiling
   per-candidate coin storage to a capacity matched to the dataset. Apple GPU compilation is unchanged.
 
--- a/src/optimization/gpu/mps_kernel.py
+++ b/src/optimization/gpu/mps_kernel.py
@@ -7,7 +7,9 @@ import time
 import numpy as np
 import torch
 
-from optimization.gpu.runtime import gpu_device, compile_shader, synchronize
+from optimization.gpu.runtime import (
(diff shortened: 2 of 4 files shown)


enarjord (commented):
No actionable findings.

Reviewed base and effective merge base `1f8f52b69facb1df9cd61f3e4454a86ce136e942`, head `e18539a8d3cc09b43d4def987e6d046982fefa7e`.

The CUDA adapter launches on Torch's current stream, matching the recorded completion event. All three unprofiled runner paths wait before decoding; profiling and temporal replay retain the existing device-wide barrier. The 0.5-second active window preserves short-kernel behavior before cooperative polling begins. Event errors propagate directly. Apple selection returns before creating a CUDA event, and its existing synchronization call is unchanged. No shader, precision, dispatch, metric, strategy, or checkpoint identity change is introduced.

Validation: 1,875 NVIDIA GPU/optimizer tests passed, 77 skipped; exact output equality in paired EMA/TM single- and multi-coin benchmarks; bounded CLI integration completed. Apple routing was tested, but Apple hardware tests were not rerun for this change. CI and independent review are separate merge gates still in progress.

Decision: semantic review clear for this exact head. This author COMMENT is not formal GitHub approval.

reviewed by GPT-6 / Codex

PR #1736, opened 2026-09-13
Fix live coin-HSL stalls from discarded fill history

enarjord:
Live coin-HSL boundary checks read the entire fill cache even when canonical startup replay has proved a narrower retained history window. Incomplete or already discarded cached episodes can therefore cause repeated reconstruction or indefinitely defer otherwise valid current trading.

Retain the initialized replay lower bound and use it only while the canonical coin-always scope remains provable. Resolve tied reset cohorts before filtering, retain late fills within the initialized window, and restore strict checking for uncertain scope evidence or threshold/never restart policies.

Validation: 487 focused HSL, risk-input recovery, and fill-history tests passed with a rebuilt, source-verified Rust extension; seven targeted delayed-boundary/initializer tests passed after enabling the retained-window regression. AI documentation checks and git diff --check passed. Tests use synthetic fixtures; no private configuration or operational data is included.


enarjord:
Reviewed head f75211df6927300ca413141c80fdc8e50499489d against base and merge base 9945eb578399c8e562c6414adbd7d9341589dc02. No actionable findings remain.

The retained boundary comes from successful canonical reconstruction, never advances with the clock, and only filters while the current coin-always scope remains provable. Tied reset cohorts retain full-tape ordering evidence before filtering. Uncertain held evidence, recent flat-scope cooldown ownership, and threshold/never policies retain strict behavior. Delayed-boundary regressions exercise replacement replay, and the rebuilt-extension focused suite passes.

Python 3.12, Python 3.14, and Rust CI passed for this head. The external Codex review completed for f75211d with no findings and a positive reaction. This author review comment is not a formal GitHub approval.

reviewed by GPT-6 / Codex

PR #1737, opened 2026-09-13
Honor shared KuCoin client timeout and credential configuration

enarjord:
KuCoin's custom broker client setup bypassed the shared CCXT configuration builder. Both clients therefore retained CCXT's 10-second request timeout, ignored explicit timeout overrides, and required legacy credential names.

Use the shared builder before installing the existing mandatory futures broker options. This gives KuCoin the common 30-second default and preserves explicit overrides and native credential aliases without changing broker signing or websocket retry policy.

Validation: four new session-construction regressions reproduced the previous failures. All 124 KuCoin, forager-websocket, and shared-client tests pass, including signed broker headers for both REST and Pro clients. The loaded Rust extension was source-verified; git diff --check passed.


enarjord:
Completed author review of head `f860b4e6349d962d2c1707a9ab6b52cae48b0ff4` against base and merge base `8a5f01fdb8faede471a77e27a0120150f1aa4d15`.

The KuCoin session factory now uses the shared configuration builder, retaining the default request timeout, explicit operator overrides, and both supported credential spellings. KuCoin's mandatory futures broker options and signing remain in place. The change does not alter websocket retry policy.

Validation: 124 focused connector and candle tests passed, including actual REST/Pro client construction and broker-signature assertions. The new timeout and credential cases fail against the previous implementation. Python 3.12, Python 3.14, and Rust CI are green on this head. The external Codex review completed without findings. No unresolved review findings remain.

This is an author review record, not a formal GitHub approval.

Reviewed by GPT-6 / Codex.

PR #1743, opened 2026-09-17
Reduce GPU suite memory use with shared arrays and lossless CUDA packing

enarjord:
GPU optimizer suites could allocate a full candle copy for every scenario even when all scenarios selected the same coins in the same order. Keep those selections as views; retain ordinary copying for actual subsets or reordered columns.

Also pack CUDA multicoin minimum-quantity comparison relations (-1, 0, +1) as signed bytes instead of int32, saving three bytes per candle per coin on both the host and GPU. CUDA lowering and argument checks enforce the matching ABI. Integer promotion preserves the comparison values; Apple GPU packing and memory safety thresholds remain unchanged.

Validation:
- GPU backend, CUDA, candle-input and prepared-data regression selection passed on NVIDIA CUDA (705 tests before the additional unspecialized-ABI case).
- Final signed-byte ABI and minimum-order/market-unstuck boundary selection: 14 passed.
- Deterministic 3- and 28-coin EMA and trailing-martingale cases: all 78 returned tensor outputs matched the pre-change CUDA results exactly for each case.
- Host-only packing test checks identical values and unchanged Apple int32 layout; suite tests cover full selections, subsets and reordered columns.
- No Rust strategy changes. Apple GPU hardware execution was not performed.


enarjord:
@codex review

enarjord:
Completed semantic review of the complete change and its input-packing and kernel consumers. No actionable findings remain. Full ordered suite selections preserve data and contiguous layout; reordered/subset selections retain their existing copy behavior. CUDA signed-byte relations preserve all three values with integer promotion and enforce the matching tensor dtype at launch. Apple packing and memory safety thresholds remain unchanged.

Validation: 705 focused tests passed, followed by 14 final CUDA ABI and boundary checks; four before/after CUDA cases produced bit-identical outputs. Python 3.12, Python 3.14, and Rust CI all passed. Independent Codex review found no major issues at this head. Apple GPU execution was not performed.

Reviewer: Codex. Decision: no actionable findings; ready to merge. This is an author sign-off, not a formal GitHub approval.
Base: 284e5de1f093341d3fa2ca5e4dcfc29b064f9c38
Head: 267817ebf453166dca60043e5612ab468cd5c77b
Effective merge base: 284e5de1f093341d3fa2ca5e4dcfc29b064f9c38

reviewed by Codex

PR #1757, opened 2026-09-19
Batch compatible Apple GPU optimizer suite scenarios

enarjord:
Small survivor populations in GPU successive halving were replayed separately for each Apple GPU suite scenario. Enable the existing compatibility-checked scenario batching for single-side multicoin Trailing Martingale on MPS, keeping scenario parameter defaults, output ordering, reducers, exact validation, and existing dispatch/work limits.

Add actual-device parity coverage for both sides, partial/full history, differing scenario defaults, and candidate/temporal chunk boundaries. A bounded synthetic benchmark compares separate versus combined batches, alternates measurement order, and requires exactly equal metrics.

Reproduce the synthetic comparison:
```bash
PYTHONPATH=src python -m tools.gpu_suite_benchmark
```
On Apple M3, the default synthetic fixture measured 48.37 seconds separately versus 11.99 seconds batched (median of two warm runs, 4.03x), with identical metrics. This measures proxy replay, not end-to-end optimizer throughput.

Validation:
- Optimizer tests: 1,740 passed and 891 device-dependent/optional cases skipped in the sandbox; the prepared-data module passed all 54 tests with POSIX shared-memory access after sandbox-only setup errors.
- 37 focused actual-Metal checks passed across suite batching, scenario result parity, recent-history views, temporal replay, and interruption behavior, using a rebuilt and fingerprint-verified Rust extension.
- End-to-end optimizer CLI smoke with synthetic cached candles and networking denied: 16 generations, 384 proxy evaluations, and all 32 exact Rust validations completed; results and checkpoint persisted successfully.
- AI documentation and generated event registry checks passed (existing documentation-size warnings only).
- `git diff --check` passed.


enarjord:
@codex review

Please review the current head after integrating master. The integration only resolves a changelog conflict; the target-relative GPU production, test, benchmark, and documentation diff is byte-identical to the original PR head. Focus on MPS scenario compatibility, materialized parameter defaults, partial-history windows, dispatch bounds, and preservation of exact validation. Actual-Metal parity, the offline optimizer CLI smoke, and synthetic A/B measurements passed; 297 focused checks passed again after integration.

PR #1788, opened 2026-09-21
Reduce revised HSL minute-price projection overhead

enarjord:
Reduce historical minute-price projection overhead in staged revised HSL. Candidate prices are collected in contiguous storage, sorted by timestamp and resolution, then reduced to the finest undisputed source. A monotonic cursor fills the minute grid without repeated tree lookups. Source eligibility, conflict detection, coarse interpolation, carried prices, timestamps and diagnostics are unchanged; no prior trading decision is cached.

Validation: all 352 Rust tests and default-feature test-target compilation pass. The rebuilt, source-verified extension passed the 1,106-test revised-HSL suite, including 80 new randomized independent-reference projection cases; 198 price/runtime tests were repeated after carrying the optimization onto the latest integration fixes. A 1,000-case before/after differential corpus, including conflicting and irregular-timestamp candles, produced identical complete outputs.

Five-run synthetic 90-day minute-grid benchmarks measured median projection time of approximately 55 ms before and 30 ms after; complete output hashes matched. This is a bounded projection improvement, not completion of the separate end-to-end admission-throughput gate. Public revised activation remains guarded and legacy remains the default.


enarjord:
Final integration check: independently cleared head `15a57c7984b5a61f6151279cc816d989c95d6813` was mechanically rebased to `0501a6d493a778236ad418fa1877320eea8a8f71` after #1786 merged. Target and effective merge base are `f32fe2d7b73255ec227ef649846cb25e639e7c18`. The complete target-relative patches are byte-identical and range-diff reports the commit unchanged. Direct inspection confirms unchanged production, regression and contract behavior. Prior semantic clearance therefore carries under the mechanical-base-update rule. Current-head CI run 35597654687 passed Python 3.12, Python 3.14 and Rust. Native source verification, 352 Rust tests, 1,106 revised tests, 198 repeated price/runtime tests and the 1,000-case differential corpus remain applicable. No findings remain.

reviewed by Codex (author mechanical integration check)

PR #1791, opened 2026-09-21
Enable revised HSL for backtests and CPU optimization

enarjord:
Allow explicit `live.hsl_engine=revised` selection through the public backtest and CPU optimizer entry points for coin, pside and unified modes. They now dispatch to the existing shared revised Rust simulator without test-local guard overrides. Legacy remains the default; live revised startup and the unsupported GPU backend still reject before activation.

Add offline integration coverage through actual backtest payloads, CLI parsing, RED/panic/cooldown reports, threshold-sensitive CPU fitness, scenario overrides, evaluator serialization and both DEAP and pymoo checkpoint resume. Tests deny IP networking in the parent and spawned workers while allowing local multiprocessing IPC. Existing schema, metric topology, saved-fitness and config-export contracts remain unchanged.

Validation: all 176 selected offline-runtime, backtest-config, shared-config and optimizer-contract tests pass with the source-verified Rust extension. This includes all eleven actual-runtime cases and the retained live-startup-before-credentials guard. Docs/registry and diff checks pass. No authenticated exchange access or live bot testing is part of this change.

Documentation distinguishes simulation availability from the still-guarded live path, explains CPU/GPU support and retains explicit migration and legacy rollback requirements. Selecting revised HSL does not itself disable public market-data downloads; offline runs still require `backtest.offline=true` and complete cached data.


enarjord:
Final author integration check: current head `eaf3e5091f84135831a7f76181846532abdcdcbc`, target and effective merge base `978c270616e98fea9412ac13bfb4ea16e73f33c5`. Independent Codex review completed for this head and signalled clearance with a PR thumbs-up; there are no review findings. CI run 35600910030 passed Python 3.12, Python 3.14 and Rust. The 176 selected source-verified offline/runtime/config/optimizer checks and six documentation tests pass, including real DEAP/pymoo resume. The reviewed diff enables simulation only, retains the live guard and legacy default, and contains only public synthetic fixtures and documentation. Branch is clean and mergeable.

reviewed by Codex (author final integration check)

PR #1794, opened 2026-09-21
Use authoritative flat positions for revised HSL cooldown recovery

enarjord:
Fresh exchange positions can show a scope is flat while delayed or incomplete fills still reconstruct exposure. Revised HSL now accepts that current flatness and uses the latest retained causal fill timestamp in the scope when the final close timestamp cannot be reconstructed. Repeated observations do not renew cooldown, and a delayed closing fill replaces the estimate with its actual timestamp.

All selected positions must be individually zero; opposing exposure cannot net to flat. The estimate is diagnostic, stays inside the configured lookback, and uses no persisted lifecycle state. Empty history creates no cooldown. The existing finite-window `never` policy and fresh-current-input requirements remain in force. Legacy behavior and the production default are unchanged.

Validation: 354 Rust tests and default-feature test compilation; rebuilt/source-verified Python extension; full revised/reference Python suite (1,555 cases), plus six added live-owner admission cases; 42 focused current-flat regressions across coin, pside and unified, both position sides, delayed fill repair, repeated observations, fresh-owner reconstruction, actual offline fake-exchange fills, stale inputs, residual exposure and lookback expiry. Integrated legacy flat-scope regressions and documentation checks pass. No live exchange tests.


enarjord:
@codex review

Please independently review current head 0174e5d1c0af6b68516ff18b54662f2b34ba7579 against master f6667a3dc2a63d32b1364cad556e2c7feb5b70d6. Focus on scope isolation, causal retained timestamps, missing final fills, cooldown/restart parity and delayed evidence correction. The explicit policy is that fresh current positions authorize flatness even when the history cannot reconstruct the closing fill; only its timestamp is estimated. Return actionable findings or explicit no-findings clearance for this head.

PR #1796, opened 2026-09-21
Fix current equity reporting for revised HSL live runs

enarjord:
Revised HSL refreshes authoritative account state without the legacy balance callback, leaving account and health equity at the constructor's near-zero placeholder. Derive these passive readings from fresh committed balance/positions and cached held-symbol quotes instead. Missing, stale, invalid, or pending inputs report null equity; zero and negative equity remain valid observations.

This introduces no exchange I/O, changes no risk or order decisions, and preserves legacy reporting. The dashboard and TUI already display null equity as unavailable.

Validation: 328 focused Python tests pass, including actual offline fake-live protective execution in all three signal modes and both sides. Eleven new regressions fail on the original implementation. Source-verified Rust extension is unchanged. Documentation and event-registry checks pass. The fix was also verified during the authorized live validation before opening this PR; operational data is excluded.


enarjord:
@codex review

Please review head 68532cf8e1b5f99665643b95d1fd5bd6f16b7494, especially passive reporting isolation, nullable-equity consumer compatibility, account/quote freshness, and preservation of legacy behavior. Focused regressions and the actual fake-live executor tests pass; source-verified native code is unchanged.

PR #1802, opened 2026-09-22
fix(cli): expose HSL engine selection for backtests and optimization

enarjord:
Backtest and optimizer CLI projection omitted `live.hsl_engine`, so `--live.hsl_engine revised` was rejected despite JSON selection and Rust backtest dispatch supporting both engines.

Register the shared selector and expose `--hsl-engine legacy|revised`, with dotted and underscore aliases. Preserve the configured engine when no override is given and the legacy default when omitted from config. This changes CLI plumbing only.

Validation: 321 focused config, CLI and revised-backtest tests passed with a source-verified native extension; actual backtest CLI smoke reached config loading with both engines without fetching market data; backtest/optimizer help and AI documentation checks passed (two existing documentation-size warnings). Regression coverage exercises command projection, all three aliases, both engines and canonical config loading across live, backtest and optimize.


enarjord:
@codex review

PR #1815, opened 2026-09-23
perf(hsl): speed up backtest reports and optimizer metrics

enarjord:
Revised HSL standalone backtests spend substantial time materializing per-minute diagnostic samples as a serde JSON tree and then rebuilding that tree as Python objects. Convert the trace directly to independent Python rows, sharing immutable field names, while preserving the complete versioned report. Metrics-only evaluations also avoid constructing discarded sample rows.

The analysis path now selects and sorts only the worst 1% tail (in the same ascending summation order) and skips EMA suffix statistics that are not returned. HSL signal, permission, execution, and report schemas are unchanged.

Validation:
- 360 Rust tests; default-feature test-target compilation; rebuilt, source-verified Python extension.
- 167 Python integration tests, including report persistence/plot consumers, real CPU optimizer DEAP/Pymoo CLI and resume, and Mac Metal single/multi-coin optimization in coin/pside/unified modes.
- Complete serde-reference transport equality and independent mutable-row tests; percentile output matches full sorting bit-for-bit, including ties, extreme magnitudes and nonfinite fallback.
- Public synthetic benchmark: `PYTHONPATH=src python tests/hsl_revised_backtest_benchmark.py --minutes 100000 --lookback-days 90 --runs 3`, with and without `--detailed`. Using the same updated harness and dependency lockfile on both revisions, median detailed native call: 0.4207s → 0.2284s; metrics-only: 0.1754s → 0.1726s. Complete result digests match before/after in both modes. These timings exclude external artifact writing and plot rendering.
- AI documentation check and `git diff --check` pass.


enarjord:
@codex review

Please independently review the current head f07232811 against master. Focus on complete report-schema/value preservation with direct Python conversion, mutable row/list independence, exact percentile summation including ties/nonfinite inputs, and whether skipped suffix EMA statistics contribute to any returned metric. Also check CPU/GPU optimizer integration and legacy result preservation. This changes reporting/analysis only, not HSL trading decisions. Please give an explicit current-head verdict after reviewing the complete diff.

PR #1816, opened 2026-09-23
perf(hsl): make per-minute backtest diagnostics opt-in

enarjord:
Revised backtests currently retain and convert a per-minute diagnostic row for every observed HSL scope, even when those traces are not needed. Make that data opt-in with `backtest.hsl_detailed_report=true` (CLI: `--backtest.hsl_detailed_report true`). Normal backtests still save effective policies, summaries and ordered RED/flat/restart events; full HSL drawdown plots require the opt-in. Metrics-only optimizer evaluations omit both sample and event lists regardless of the option.

The reporting switch does not change reconstruction, decisions, fills, equity or analysis metrics. Event capture is independent of sample capture, including same-timestamp transitions and liquidation-proven flattening. Update configuration validation, plotting guidance, documentation and the public offline benchmark accordingly.

Validation:
- 361 Rust tests; default-feature Rust test compilation.
- Rebuilt and source-verified native extension; 269 focused Python tests, including report/metric/trading parity across coin, pside and unified modes, configuration roundtrips, both CLI report modes, and offline CPU optimizer/resume coverage.
- Documentation and generated-registry checks passed (existing size warnings only).
- Public offline benchmark, 100,000 minutes / 90-day lookback / three runs: median 0.174 s with default event-only reporting versus 0.222 s with full per-minute reporting. Reproduce with `PYTHONPATH=src python tests/hsl_revised_backtest_benchmark.py --minutes 100000 --lookback-days 90 --runs 3 --detailed`, adding `--hsl-detailed-report` for the full trace. Exact trading/analysis/event parity is covered separately; complete result hashes intentionally differ because the sample list is omitted.


enarjord:
@codex review. Please review the current head b4273c251 for reporting-only isolation, retained lifecycle transitions, configuration/CLI defaults, optimizer artifact suppression, and report/plot consumers. Trading outputs and analysis metrics must remain unchanged.

enarjord:
@codex review

Please review current head 15772b8966f078316d612aab25d6e462240ef602. Since the cleared b4273c2517, the sole change adds the canonical false reporting default to two manually assembled backtest config fixtures. Production code is unchanged. Both CI Python jobs caught those six fixture failures. Python 3.14 additionally had one asynchronous fake-live replay comparison failure; the same test passed in Python 3.12 CI and the complete local fake-cycle suite passed without modifying its assertions or live code. The affected fixture/fake-cycle suites now pass all 90 tests locally. Fresh CI is required before merge.

PR #1817, opened 2026-09-23
chore(hsl): compile full-history reference wrapper only in tests

enarjord:
`revised_hsl_inputs_at` is a full-lookback reference wrapper used only by parity tests; production captures inputs through `revised_hsl_inputs_at_clipped`. Gate the wrapper with `#[cfg(test)]` so release builds no longer emit an unused-method warning while preserving test coverage. No trading behavior changes.

Validation: warning-free `cargo check --lib`, `cargo check --tests`, 361 Rust tests, a rebuilt/source-verified extension, and 49 native backtest/reporting Python tests.


enarjord:
@codex review. Please confirm the helper is used only by test callers and that gating it with cfg(test) preserves production compilation and parity-test coverage.

PR #1819, opened 2026-09-23
fix(bitunix): scope live quote refreshes to requested symbols

enarjord:
Bitunix's automatic live quote route requested every active market. Its websocket reader waits up to eight seconds for that entire set, so an unrelated silent market could delay an already-available held-symbol quote beyond the revised HSL quote reader's deadline.

Default Bitunix snapshots to targeted reads, preserving websocket receipt freshness, the bounded depth fallback, and explicit bulk overrides. Protective and ordinary planning now wait only for the symbols they requested.

Validation: 524 connector, shared-snapshot and revised-HSL tests passed, including six offline fake-live panic executions across coin, pside and unified modes. Three new regressions fail on the base and pass with this change. The real connector/provider/protective-reader regression repeatedly expires the snapshot cache while another market never supplies a ticker. Documentation and generated-registry checks pass. Source-verified live verification confirmed continuing HSL evaluation and ordinary planning through the targeted route; no live panic was forced.


enarjord:
@codex review current head 3e931f152f1f90b4697f86e4bd3f2df433603706. Please check scoped quote readiness, explicit overrides, bounded fallback and the connector/provider/revised-owner regressions. The default routing change must not weaken freshness or alter strategy decisions.

